PR #13, opened 2019-04-29
fix: use find-my-way.find() for pure routing

chrisdickinson:
Remove the router's control over the request/response cycle. This solves the problem where 404's were being written to the response twice, and probably future problems as well.

Implications: 

- `this` in a handler is no longer the request context.
- The first param of a handler will be the request context, the second will be the route parameters, the third will be the mysterious "store" object alluded to in the find-my-way docs:

```javascript
router.get('/defrobnicator/:problemFrobs', frobRemovalHandler)

async function frobRemovalHandler (context, { problemFrobs }) {
}
```

### DIFF
--- a/src/registry/lib/router.js
+++ b/src/registry/lib/router.js
@@ -1,8 +1,12 @@
+'use strict';
+
 // This router is an *even simpler* version of micro-fork.
 // It uses a context object instead of a req/res pair.
 // See https://github.com/amio/micro-fork
 
+const response = require('./response');
 const fmw = require('find-my-way');
+const URL = require('url');
 
 function router(options) {
   const wayfinder = fmw(options);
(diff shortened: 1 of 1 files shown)


delvedor:
Hello! `find-my-way` author here :)
Thank you for using this package, and thank you for creating this fantastic project!
If I can help you with something, please let me know!!

Regarding the "mysterious" `store` object, it's a simple parameter that you can ignore if you don't need it. You can configure it during the route creation, for example
```js
router.on('GET', '/', handler, { hello: 'world' })

function handler (req, res, params, store) {
  // you can read the store object here
}
```
It comes conveniently if you need to pass around a custom context for every route, for example, we are using it in Fastify for handling the encapsulation model, where we create a set of functionalities and steps that a specific route must have or run.

PR #103, opened 2019-06-01
The most extremely basic website

trodrigues:
Assumes nothing about what might be necessary in the future, just adds some basic sample content from the repository and some very basic styles I lifted from my own website.

### DIFF
--- a/web/styles.css
+++ b/web/styles.css
@@ -1,0 +1,49 @@
+header h1 {
+  padding: 0;
+  margin: 0 0 0.3em;
+  color: salmon;
+}
+
+header p {
+  margin: 0;
+  padding: 0;
+  color: lightslategray;
+}
+
+nav ul {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+
+nav li {
+  display: inline;
+}
+
+nav li a {
+  text-decoration: none;
+}
+
+nav li a:hover {
+  text-decoration: underline;
+}
+
+nav li:after,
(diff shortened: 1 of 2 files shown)


web/styles.css
```diff
@@ -0,0 +1,105 @@
...
+  --link-visited: hsl(14, 51%, 74%);
+  --separator: rgb(150, 150, 150);
+}
+
+body {
+  font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
+  width: 800px;
+  margin: 0 auto;
+  padding-top: 2.5em;
+  padding-left: 0
```
lucask42: missing semi

web/styles.css
```diff
@@ -0,0 +1,105 @@
...
+.vertical-separator:after {
+  content: '/';
+  color: var(--separator);
+  margin-left: 0.7em;
+  margin-right: 0.4em;
+}
+
+nav li:last-child:after,
+.vertical-separator:last-child:after {
+  content: ''
```
lucask42: missing semi (technically is optional if it is the last rule in a block)

Not setting margin-left and margin-right to 0 here leaves them on the page.  Just a nit pick

trodrigues: Thanks for the comments! This was something I did super quickly on my ipad and I easily missed these 😅

PR #123, opened 2019-06-03
[feat] Set up system for locale-specific help

toddself:
Provide a way to determine the locale of the user and load in a
localized help file, defaulting to english.

Add example help file from #111

### DIFF
--- a/cli/lib/commands/login.js
+++ b/cli/lib/commands/login.js
@@ -14,7 +14,6 @@ const loginOpts = figgy({
 });
 
 async function login(opts) {
-  console.log(opts);
   opts = loginOpts(opts);
 
   const { username, token } = await profile.loginWeb(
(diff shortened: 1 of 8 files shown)


zacanger:
@toddself your build may be broken because I had some bugs in a commit, if you pull from upstream master I think that'll fix that.

cli/lib/commands/login.js
```diff
@@ -14,6 +14,7 @@ const loginOpts = figgy({
 });
 
 async function login(opts) {
+  console.log(opts);
```
chrisdickinson: EX-TERM-IN-ATE

toddself:
thanks @zacanger i rebased off master.  @chrisdickinson that console.log has been taken out!

chrisdickinson:
Yay, awesome work. Merging now!

PR #119, opened 2019-06-02
Local dev docker setup

zacanger:
Related: #116. This fleshes out the docker-compose file and enables one-command local development.

PR #125, opened 2019-06-03
Fix tests and docker-compose

zacanger:
#119 broke registry tests. I believe this fixes it. Would one of y'all be able to enable Circle builds on PRs?


registry/test/utils/postgres.js
```diff
@@ -7,7 +7,12 @@ const { Client } = require('pg');
 
 function providePostgres(to) {
   return async function(...args) {
-    const client = new Client();
+    const client = new Client({
+      user: process.env.PGUSER,
+      database: process.env.PGDATABASE,
+      port: 5432,
+      host: process.env.POSTGRES_HOST
```
chrisdickinson: A-ha! So! node-pg (and many other postgres tools) will use `PGUSER` / `PGDATABASE` automatically if they're set in the environment. If we switch from `POSTGRES_HOST` to `PGHOST` the original code _should_ work.

zacanger: Oh good to know! It looks like the env var CJ added in Circle fixed it, but I changed the names anyway in case you'd rather have that in code.

zacanger:
Did you set it to `localhost`? I use Circle for some projects but I'm not an expert, not sure what it does with docker networking when you have multiple containers.

zacanger:
Added in the circle config, along with EXTERNAL_HOST.

docker-compose.yml
```diff
@@ -30,15 +30,15 @@ services:
   registry:
     image: mhart/alpine-node:12
     volumes:
-      - ./registry:/registry
-    working_dir: /registry
+      - ./services/:/services
```
zacanger: Mounting all of `services` because of the boltzmann `file:` dependency

ceejbot: TY!

PR #126, opened 2019-06-03
Change the wording of "reverse isms" in the COC

mvasilkov:
I realize that the current COC is boilerplate, but the amended line basically encourages/deems acceptable discriminating against white people (the "reverse racism" part). I think it is appropriate to reword it to explicitly condemn racism and sexism as a whole, without the otherwise implied cherry-picking of racial/gender groups that are allowed to be discriminated against.

An example of a solid COC is actually linked from the current boilerplate:
- https://www.recurse.com/manual
- https://www.recurse.com/code-of-conduct


chrisdickinson:
The current code of conduct was specifically selected and is not up for debate.

Issue #116, opened 2019-06-02
Kubernetes support


zacanger:
Might be nice to have a helm chart or k8s files rather than using `docker-compose` and/or manually starting, for deployment. I could work on this if you think it'd be valuable.

* Fully dockerized local dev setup: #119
* k8s setup for easy prod standup: #204

ceejbot:
I would *love*  this. We need easy start-up both for development and for easy stand-up by  administrators.

toddself:
@zacanger i'd love to help out with this if you need any assistance!

zacanger:
Definitely! I'm still learning Kubernetes so any help is appreciated. I got a rough start [here](https://github.com/zacanger/entropic/commit/23309c90c3554ac26ca77e389d23c4ac5453ca60) but took a break from it due to minikube using literally all of my laptop's resources.

toddself:
awesome i can give this some testing!

also we might consider using something like [pulumi](https://www.pulumi.com/kubernetes/) or another tool rather than having to manage yaml by hand. i ... dislike yaml a lot.

zacanger:
Some tool to enable folks to work with anything other than yaml would be amazing, I think. I spend way too much time fiddling with indentation. Even just using a js to yaml package might be a pretty good solution. @ceejbot what do you think?

zacanger:
@toddself I made some pretty big changes over here https://github.com/zacanger/entropic/tree/feature/k8s/misc/k8s -- now building from plain JS objects, and spent some time cleaning it up.

toddself:
oh awesome! i've been heads down on my day job (and writing a Package.toml RFC) but will check this out tonight!

krainboltgreene:
I know quite a bit about this surface of application development and I would love to do a pair review to transfer the knowledge.

alexindigo:
@zacanger is there anything left to tackle? I spent last year and half doing exactly this, so I'd like to chip in. :)

zacanger:
@alexindigo thanks! I think #204 is good to go, but definitely welcome more eyes on it!

alexindigo:
Thanks @zacanger, I see it's k8s related changes. Are you interested in local docker development?

alexindigo:
And you didn't go with helm just because of yaml or didn't want to have go templates in the stack?

zacanger:
Local docker env was set up in #119, so now running locally should be as simple as `npm i` and `docker-compose up`.

[...]

PR #130, opened 2019-06-03
Add registry Docker build

zacanger:
A prerequisite for #116 is probably having the registry (and future services) already containerized and on some image registry. This adds a Docker build and makes some minor modifications.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,1 +1,9 @@
 node_modules
+.env
+.nyc_output
+coverage
+legacy-cache
+allowed-list
+.DS_Store
+*~
+*.swp
--- a/cli/lib/canonicalize-spec.js
+++ b/cli/lib/canonicalize-spec.js
@@ -35,7 +35,7 @@ function parsePackageSpec(input, defaultHost) {
   const [namespace, host = defaultHost] = namespacehost.split('@');
   const [namePartialEnc, range = 'latest'] = namerange.split('@');
 
-  const name = decodeURIComponent(namePartialEnc)
+  const name = decodeURIComponent(namePartialEnc);
   return {
     canonical: `${namespace}@${host}/${encodeURIComponent(name)}`,
     host,
--- a/misc/build-registry-image.sh
+++ b/misc/build-registry-image.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+version=$(git descibe --exact 2>/dev/null || git rev-parse --short=10 HEAD)
+docker build -t entropic/registry:$version -f services/registry.dockerfile services
--- a/services/registry.dockerfile
+++ b/services/registry.dockerfile
@@ -1,0 +1,21 @@
+FROM mhart/alpine-node:12
+RUN addgroup -g 1000 -S node && \
+    adduser -u 1000 -S node -G node
+WORKDIR /services
+COPY --chown=node:node registry registry
+COPY --chown=node:node common common
+WORKDIR /services/registry
+RUN npm install --quiet
+USER node
+ENV NODE_ENV=production \
+    TERM=linux \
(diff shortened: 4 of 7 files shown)


misc/build-registry-image.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+version=$(git descibe --exact 2>/dev/null || git rev-parse --short=1- HEAD)
```
zacanger: `entropic/registry:vX.X.X` or `entropic/registry:XXXXXXXXXX`. Since there's no Docker repo yet I can change this if need be.

services/registry/healthcheck.js
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env node
+
+// Use Node http rather than installing curl
+// to reduce container size
+const request = require('http')
```
zacanger: Simple healthcheck script

services/registry/server.js
```diff
@@ -31,6 +31,19 @@ const myMiddles = [
   require('./middleware/object-store')
 ];
 
-const server = boltzmann.make(router, myMiddles);
-server.listen(process.env.PORT, '0.0.0.0');
-logger.info(`listening on port: ${process.env.PORT}`);
+const main = () => {
```
zacanger: This is just a little modification to make closing the server a little more graceful in Docker.

services/registry.dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM mhart/alpine-node:12
```
zacanger: Working from one directory above `/registry` because we can't add files from outside the Docker context (the `common` directory).

services/registry.dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN addgroup -g 1000 -S node && \
```
zacanger: If the process needs to write to the disk, using a limited user like this might not work.

cli/lib/canonicalize-spec.js
```diff
@@ -35,7 +35,7 @@ function parsePackageSpec(input, defaultHost) {
   const [namespace, host = defaultHost] = namespacehost.split('@');
   const [namePartialEnc, range = 'latest'] = namerange.split('@');
 
-  const name = decodeURIComponent(namePartialEnc)
+  const name = decodeURIComponent(namePartialEnc);
```
zacanger: This was from running Prettier.

.dockerignore
```diff
@@ -1 +1,9 @@
 node_modules
+.env
+.nyc_output
+coverage
+legacy-cache
+allowed-list
+.DS_Store
+*~
+*.swp
```
ceejbot: So many ignore files with nearly-identical contents...

misc/build-registry-image.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -e
+
+version=$(git descibe --exact 2>/dev/null || git rev-parse --short=10 HEAD)
+docker build -t entropic/registry:$version -f services/registry.dockerfile services
```
ceejbot: I almost twitched and said, oh hey, a job for a makefile, but not yet, not yet.

services/registry.dockerfile
ceejbot: This process definitely needs to write to *some* volume right now. (Using an object store for storage is on the list, but some users will never want to do that.)

services/registry/server.js
ceejbot: Aha, excellent.

.dockerignore
zacanger: I know, maybe someone should invent the 16th standard...

PR #131, opened 2019-06-03
Add basic issue and PR templates

zacanger:
Having some basic fields to fill out for issues and PRs is something I find useful on my own projects and at work (we require PR templates on all repos, and it really helps!). These are the ones I use on my FOSS stuff as a good baseline.

Also adds an editorconfig and links to some files.


.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
```
olingern: @zacanger did you mean to add this config file along with the PR templates?

zacanger:
Thanks for the suggestions @varjmes, I applied those!

Issue #83, opened 2019-05-23
write registry tests enough to get code coverage to 80%


ceejbot:
We don't have tests. It will help potential contributors contribute if we have tests. We should write some.

We should see if we can make it easier to build a test rig for the registry, or if the provider-function approach we have now is enough.

PR #132, opened 2019-06-04
Add token, user, package model tests

olingern:
I thought it would be nice to get to know the project and write some of the registry's model tests. 

Related issue #83


chrisdickinson:
Thanks for the PR! Ideally I'd like to use the `providePostgres` helper for all tests that touch the database: the goal of the helper is not just to provide the connection, but also to run all modifications to the database during a testcase in a transaction which is rolled back at test completion. This pattern (borrowed from Django) isolates test state between test cases. It'd definitely be worthwhile to look into allowing the connection to be reused between test cases, though!

olingern:
@chrisdickinson got these passing :tada: 

- Had to add the env vars to the circle ci config
- Add mocha to the eslintrc

olingern:
@chrisdickinson to the comment above:

`the goal of the helper is not just to provide the connection, but also to run all modifications to the database during a testcase in a transaction which is rolled back at test completion`

So, I read this and the `providePostgres` function as:
- Start with a clean slate every individual test
- Wrap everything in a transaction

By using `beforeEach` we're saying, _"Open a transaction every test"_ and using `afterEach` rollback the previous transaction.

I do like the `providePostgres` function, but (_currently_) it doesn't allow errors to propagate up and it's difficult to capture results inside the passed function scope and test them outside of it. My specific use case is capturing the errors that Joi / ormnomnom throws. I'm not tied to this. If there's a better way to do with within the confines of `provideRegistry`'s implementation, I can do that :slightly_smiling_face:

zacanger:
@olingern can you pull from upstream master? I think some of the things you're fixing are fixed already.

olingern:
@zacanger Sure, done. I only saw one conflict which was with d54ab504c146ef327cf5d209e3a2aded4cdfb72a, so these should be okay

olingern:
@chrisdickinson refactored per your comment above. Also moved this under services. Let me know if you'd like me to drop b5b681dae3a24258b170e5cabba15ef1f0c018ec per the conversation in #133

olingern:
Will squash / fixup these commits after review :slightly_smiling_face:

services/registry/models/token.js
```diff
@@ -53,6 +52,10 @@ module.exports = class Token {
   }
 
   static async lookupUser(value) {
+    if (!value || value.indexOf('ent_v1_') !== 0) {
```
olingern: Sanity check added here.

services/registry/models/token.js
```diff
@@ -1,6 +1,5 @@
 'use strict';
 
-const querystring = require('querystring');
```
olingern: I saw this wasn't used

services/registry/test/models/user.js
```diff
@@ -0,0 +1,173 @@
...
+        // No authentication objects should exist for this user
+        demand(remoteAuthObjects.length).to.be(0);
+      })
+    );
+
+    it(
+      'should create a remoteAuth when remoteAuth param is provided',
+      providePostgres(async () => {
+        // Kind of dirty, but we currently only need it for this test
+        let prevOAuthEnvVar = process.env.OAUTH_PASSWORD;
```
olingern: Like my comment says, it's not the best. Altering `process.env` feels weird ... but I think we would need some sort of global setup script that would load something like a `.env-test` file to avoid this

services/registry/test/models/package.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+const demand = require('must');
+const providePostgres = require('../utils/postgres');
+const Package = require('../../models/package');
+const Namespace = require('../../models/namespace');
+
+const { createUser } = require('../utils/users');
+
+async function getUserNamespace(name, email) {
```
olingern: This might should be moved depending on how "helpful" it is

.eslintrc
```diff
@@ -1,12 +1,14 @@
 {
   "env": {
     "node": true,
-    "es6": true
+    "es6": true,
+    "mocha": true
```
olingern: Added `mocha` option due to getting `described` and `it` are undefined on the CI.

services/registry/test/models/user.js
jacques: Better to use the .env-test file for this.

olingern:
Rebased master that has #154 in it, and now passing :tada: 

Cleaned this up and keeping it to only token, user, and package tests

services/registry/models/token.js
chrisdickinson: 🎉 This is a great idea! (A tiny nit, but could we switch from `indexOf` to `startsWith`?)

services/registry/test/models/package.js
```diff
@@ -0,0 +1,40 @@
...
+    active: true,
+    name: newUser.name
+  });
+}
+
+describe('Package', () => {
+  before(async () => {});
+
+  it(
+    'Given minimum, valid paramters it creates a new package ',
```
chrisdickinson: typo: `parameters`

services/registry/test/models/user.js
chrisdickinson: I don't mind messing with the environment here. I might suggest doing this bit in a separate "before / after" pair to make sure it's always run.

services/registry/test/models/user.js
```diff
@@ -0,0 +1,172 @@
...
+        const remoteAuthObjects = await Authentication.objects.filter({
+          user
+        });
+
+        // The user should now have 1 active authentication object
+        demand(remoteAuthObjects.length).to.be(1);
+        demand(remoteAuthObjects[0].active).to.be.truthy();
+
+        // reset
+        process.OAUTH_PASSWORD = prevOAuthEnvVar;
```
chrisdickinson: Typo: `process.env.OAUTH_PASSWORD` (missing `.env`)

olingern:
@chrisdickinson thanks for the review! I've been having some problems with testing the few nits, so I opened #190

services/registry/test/models/user.js
olingern: :eyes:

services/registry/test/models/package.js
olingern: Fixed!

services/registry/test/models/user.js
olingern: > I might suggest doing this bit in a separate "before / after" pair to make sure it's always run.
Done!

services/registry/models/token.js
olingern: Sure, updated

olingern:
@chrisdickinson I went through each nit and updated and also:
- Updated the `npm test` command to use a glob
- Renamed the test files to `test.spec.js` ( _this allows us to easily exclude utils instead of a complex regex_ )
- Moved the integration tests under `integration`

If those changes are :+1: I'll do some commit squashing

services/registry/test/models/user.spec.js
```diff
@@ -0,0 +1,178 @@
...
+
+    it(
+      'creates a namespace with a host for the user',
+      providePostgres(async () => {
+        const name = 'foo bar';
+        await User.signup(name, 'bar@foo.com', false);
+
+        // Fetch the user's namespace that was created during signup.
+        const userNamespace = await Namespace.objects.filter({
+          'host.id': 1, // TODO: don't hardcode?
```
chrisdickinson: I'd be tempted to add a helper function to `Host` to grab the current hostname so we can get condense all of the `process.env.EXTERNAL_HOST.replace(/https?:\/\//, '')` bits into a single `Host.current()` call. That would let us write `'host.name': Host.current()` here, for example!

(This can definitely be left for a separate PR.)

olingern: Oh, that would be nice! When I first looked at this, I had to do some logging to figure out what 1 actually meant. So a helper would be ... helpful :slightly_smiling_face:

olingern:
Alright, rebase done. Moved the previous registry tests into storage and updated accordingly. I should note that `pretest` fails on my machine. Can't connect to the postgres instance. Removing `pretest` and then running `npm t` inside of the storage service works, though

chrisdickinson:
Thanks for working on this! I'm going to plan on merging this tomorrow AM.

chrisdickinson:
Merged in 7f935831f1da132a92e964b36ef3434b92223575, thanks!

olingern:
awesome :rocket: !

PR #135, opened 2019-06-04
Split the registry into registry, web, and storage services

chrisdickinson:
Status: it's ready!


package.json
```diff
@@ -31,8 +31,8 @@
     "lint-cli": "cd cli; npm run lint",
     "lint-fix": "prettier --write '**/*.js'",
     "lint-registry": "cd services/registry; npm run lint",
-    "postinstall": "for d in cli services/registry services/workers services/common/boltzmann; do cd $d; npm i; cd -; done",
+    "postinstall": "for d in cli services/registry services/workers services/web services/common/boltzmann; do cd $d; npm i; cd -; done",
     "start": "./misc/start.sh",
```
ceejbot: Mostly I'm just thinking about how much work is ahead re-implementing run-scripts.

services/common/boltzmann/middleware/requestid.js
```diff
@@ -2,16 +2,18 @@
...
 
 function createRequestId(
   requestIdHeader = process.env.REQUEST_ID_HEADER || 'request-id'
 ) {
+  const host = os.hostname()
   return function mw(next) {
     return async function inner(context) {
       const request = context.request;
-      context.id = request.headers[requestIdHeader] || uuid.v1();
+      context.id = request.headers[requestIdHeader] || `${host}_${uuid.v1()}`;
```
ceejbot: Aha. Excellent.

services/web/NOTES
```diff
@@ -0,0 +1,19 @@
...
+        * NOTE: bring email uniqueness check into signup
+        * NOTE: err.code
+        User
+    getTokens({for: user, page: N})
+      [objects]
+    createToken({for: user, description})
+    deleteToken({for: user, valueHashes: [String]}) -> { count }
+    resolveCLISession({session, value})
+    fetchCLISession({session}) -> { description }
+
```
ceejbot: This is, I note, the minimal useful web feature set. I think it's smart of you to avoid things like package visualization for now.

services/web/handlers/auth.js
```diff
@@ -9,23 +9,20 @@ const { text } = require('micro');
...
-  fork.get('/www/login', handleCLISession(redirectAuthenticated(login))),
-  fork.get('/www/signup', redirectAuthenticated(signup)),
-  fork.post('/www/signup', redirectAuthenticated(signupAction)),
-  fork.get('/www/tokens', seasurf(redirectUnauthenticated(tokens))),
-  fork.post('/www/tokens', seasurf(redirectUnauthenticated(handleTokenAction)))
+  fork.get('/login', handleCLISession(redirectAuthenticated(login))),
+  fork.get('/signup', redirectAuthenticated(signup)),
+  fork.post('/signup', redirectAuthenticated(signupAction)),
+  fork.get('/tokens', seasurf(redirectUnauthenticated(tokens))),
+  fork.post('/tokens', seasurf(redirectUnauthenticated(handleTokenAction)))
```
ceejbot: Yeah, putting the endpoints into registry for now is good. You might consider namespacing them. I think that'll be good even after we pull things into a second service.

services/web/handlers/auth.js
```diff
@@ -359,24 +351,17 @@ async function handleTokenAction(context) {
...
-        'user.name': user,
-        'user.active': true,
-        active: true,
-        value_hash: token
-      })
-      .update({ active: false });
+    const { count } = await context.storageApi.deleteToken({for: user, valueHashes: [value_hash]})
 
-    context.session.set('banner', 'Successfully deleted 1 token.');
+    context.session.set('banner', `Successfully deleted ${count} token${count === 1 ? '' :  's'}.`);
```
ceejbot: the thin edge of the wedge with ternaries

services/web/server.js
```diff
@@ -0,0 +1,33 @@
...
+
+const logger = bole('runner');
+if (isDev()) {
+  const prettystream = bistre({ time: true });
+  prettystream.pipe(process.stdout);
+  bole.output({ level: 'debug', stream: prettystream });
+} else {
+  bole.output({ level: 'info', stream: process.stdout });
+}
+
```
ceejbot: I note that 90% of this is boilerplate. We should make a generic wrapper.

services/common/boltzmann/client/NOTES
```diff
@@ -0,0 +1,100 @@
+storageApi
```
chrisdickinson: Hey look, it's almost a complete list of everything an entropic can do!

services/common/boltzmann/client/index.js
```diff
@@ -0,0 +1,207 @@
+'use strict'
```
chrisdickinson: Both the web gateway and registry gateway can use this client to talk to storage.

services/common/boltzmann/client/index.js
```diff
@@ -0,0 +1,207 @@
...
+  constructor ({
+    url = process.env.STORAGE_API_URL,
+    requestId,
+    userAgent,
+    hostname = hostname,
+    logger
+  }) {
+    this.url = url
+    this.requestId = requestId
+    this.userAgent = `${hostname}(${userAgent})`
```
chrisdickinson: Wrap up the incoming requestId and userAgent (so we know which host is sending which user-agent where, in logs.)

Should I add `x-forwarded-for` here as well, so remote address logging works correctly in storage-api?

services/common/boltzmann/client/index.js
```diff
@@ -0,0 +1,207 @@
...
+      method,
+      headers: {
+        ...headers,
+        'user-agent': this.userAgent,
+        'request-id': this.requestId
+      },
+      body: rawBody ? rawBody : body ? JSON.stringify(body) : null
+    })
+
+    if (!response.ok) {
```
chrisdickinson: The fetch api is so nice.

services/common/boltzmann/request-handler.js
```diff
@@ -22,6 +24,15 @@ class Context {
...
     this.host = host;
+    this._parsedUrl = null;
+  }
+
+  get url () {
+    if (this._parsedUrl) {
+      return this._parsedUrl
+    }
+    this._parsedUrl = parse(this.request.url, true)
+    return this._parsedUrl
```
chrisdickinson: Be lazy about URL parsing.

services/registry/handlers/maintainers.js
```diff
@@ -16,88 +10,72 @@ module.exports = [
...
-        `${invitee} has already been invited to maintain ${namespace}@${host}/${name}.`
-      );
-    }
+  if (err) {
+    const msg = {
+      'maintainer.invite.invitee_dne': `Unknown namespace: "${invitee}".`,
+      'maintainer.invite.package_dne': `Unknown package: "${invitee}".`,
+      'maintainer.invite.already_accepted': `Namespace "${invitee}" is already a member.`,
+      'maintainer.invite.already_declined': `Namespace "${invitee}" has declined this invite.`
+    }[err.code]
```
chrisdickinson: The error mapping pattern emerges. It's almost worth making a decorator for this and throwing the error, I think. This is where we should hang i18n mapping as well, potentially.

services/registry/middleware/bearer-auth.js
```diff
@@ -34,16 +21,29 @@ function createBearerAuthMW() {
         );
       }
 
+      // getting access to the redis doesn't get you the tokens.
+      const hash = crypto
+        .createHash('sha256')
+        .update(bearer + process.env.SESSION_SECRET)
+        .digest('base64')
```
chrisdickinson: The registry now caches auth in redis for a span of time, so we hash it.

services/registry/package.json
```diff
@@ -12,13 +12,10 @@
...
-    "cls-hooked": "^4.2.2",
     "cookie": "^0.3.1",
-    "csrf": "~3.1.0",
     "culture-ships": "~1.0.0",
     "dotenv": "~8.0.0",
     "escape-html": "^1.0.3",
-    "find-my-way": "~2.0.1",
     "graceful-fs": "^4.1.15",
     "is-email-maybe": "^1.0.1",
     "markdown": "^0.5.0",
```
chrisdickinson: TODO: review all service deps and shrink to the minimum set for each service.

chrisdickinson:
This turned out to be much bigger than expected! (It's good to get this out of the way now, though, because it's just going to move _more_ cheese as time goes on.)

Things left to do:

- [ ] Finish wiring up the client
- [ ] Fill out the remaining "TODO" api endpoints from the client (things we were previously glomming together with the registry/web endpoints before, like token creation)
- [ ] Do a error enumeration pass through:
  - [ ] the registry
  - [ ] the website
  - [ ] the storage layer
- [ ] Do a logging pass through:
  - [ ] the registry
  - [ ] the website
  - [ ] the storage layer

services/common/boltzmann/client/index.js
```diff
@@ -0,0 +1,207 @@
+'use strict'
+
+const fetch = require('node-fetch')
```
olingern: You might want to use the wrapped `fetch` that was merged in #177. #183 is still open that aims to do some refactoring of the original merge

services/common/boltzmann/client/index.js
```diff
@@ -0,0 +1,207 @@
...
+const os = require('os')
+
+const hostname = os.hostname()
+
+const e = encodeURIComponent
```
olingern: What do you think of wrapping this in a function? If any future refactors need to happen it could make it easier.

`const encode = (str) => encodeURIComponent(str)`

services/common/boltzmann/client/index.js
```diff
@@ -0,0 +1,207 @@
...
+
+module.exports = class Client {
+  constructor ({
+    url = process.env.STORAGE_API_URL,
```
olingern: Should we `assert`, `throw`, or warn here if that env var isn't defined?

services/web/server.js
olingern: My go at code golfing the wrapper:

**Wrapper**
```javascript
function boltzStarter(router, middles, boltzmann) {
  return boltzmann.make(router, middles.map(middle => require(middle)));
}
```
**Usage**
```javascript
const myMiddles = [
  'boltzmann/middleware/logger',
  'boltzmann/middleware/flush-request',
  'boltzmann/middleware/requestid',
  'boltzmann/middleware/redis',
  'boltzmann/middleware/storage-api',
  './middleware/session',
];

boltzStarter(router, myMiddles, boltzmann).listen(
  process.env.PORT,
  '0.0.0.0'
);
```

services/common/boltzmann/client/index.js
```diff
@@ -185,9 +300,10 @@ module.exports = class Client {
         message = parsed.message || body
         code = parsed.code || code
       } catch {
+        ;
```
olingern: I'm sure you're probably working on this now, but commenting just in case. Empty line / no caught error

services/storage/handlers/namespaces.js
```diff
@@ -0,0 +1,253 @@
...
+    const ns = await Namespace.objects
+      .get({
+        active: true,
+        name: params.namespace,
+        'host.active': true,
+        'host.name': params.host,
+        'namespace_members.active': true,
+        'namespace_members.user_id': context.user.id
+      })
+      .catch(Namespace.objects.NotFound, () => null);
```
olingern: Could we add a log here? #202 took me a bit to track down :sweat_smile:

olingern:
Awesome work! I thought I would try to help PR review and leave some comments. I'll try to test this branch out at some point to see if I can turn up any :beetle:

services/common/boltzmann/request-handler.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+const { parse } = require('url');
```
zacanger: `url.parse` is deprecated, should this use `new URL` instead?

services/registry/handlers/maintainers.js
```diff
@@ -16,88 +10,71 @@ module.exports = [
...
-    .catch(Package.objects.NotFound, () => null);
+  const [err, maintainers] = await context.storageApi.listPackageMaintainers({
+    namespace,
+    host,
+    name,
+    bearer: context.user ? context.user.name : null,
+    page: Number(context.url.query.page) || 0,
+    status: context.url.query.status
+  }).then(
+    xs => [null, xs],
```
zacanger: I like this. Haskell influence?

services/registry/handlers/users.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+async function accept(context, { username, namespace, host }) {
+  const [err] = await context.storageApi.acceptNamespaceMembership({
+    bearer: context.user.name,
+    invitee: username,
+    namespace,
+    host
+  }).then(
+    xs => [null, xs],
```
zacanger: Would it be worth pulling this into an `either` util?

services/storage/handlers/packages.js
```diff
@@ -0,0 +1,509 @@
+'use strict';
+
+const { Response } = require('node-fetch');
+const { markdown } = require('markdown');
```
zacanger: This particular markdown lib is very abandoned. It might make sense to use something like markdown-it, which is both maintained and super pluggable.

services/storage/handlers/packages.js
```diff
@@ -0,0 +1,509 @@
...
+    ++filecount;
+    part.on('error', err => {
+      validationError = err;
+    });
+
+    const filename =
+      './' + decodeURIComponent(String(part.filename)).replace(/^\/+/g, '');
+    formdata.files[filename] = context.storage.add(part);
+
+    if (/^\.\/package\/readme(\.(md|markdown))?/i.test(filename)) {
```
zacanger: `.mkd` is also somewhat common

services/storage/handlers/packages.js
```diff
@@ -0,0 +1,509 @@
...
+
+  if (filecount > MAX_FILES) {
+    return response.error(
+      `Exceeded maximum number of files in a version.`,
+      400
+    );
+  }
+
+  await Promise.all(
+    Object.keys(formdata.files).map(filename => {
```
zacanger: `Object.values` might reduce some code here

services/storage/server.js
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env node
+'use strict';
+
+require('dotenv').config();
+
+const boltzmann = require('boltzmann');
```
zacanger: I'd be happy to put in a PR to pull out some of this boilerplate, it'll probably become pretty common.

services/web/handlers/index.js
```diff
@@ -0,0 +1,31 @@
...
+async function ping() {
+  return response.text(ship);
+}
+
+async function homepage() {
+  return response.html(`
+    <!doctype html>
+    <html>
+      <body>
+        <h1><marquee>WELCOME TO ENTROPIC</marquee></h1>
```
zacanger: `WELCOME TO 1997`

services/web/server.js
zacanger: Oh, I commented above but it looks like y'all are already on it.

services/common/boltzmann/client/index.js
chrisdickinson: Ah, this is on the server side!

services/storage/handlers/packages.js
chrisdickinson: I was waiting to use `marky-markdown`, which is a Known Good package for rendering markdown; _however_ it brings along oniguruma so I wanted to wait until we had separated package publishes out into a worker process.

services/storage/handlers/packages.js
chrisdickinson: great catch!

services/storage/handlers/packages.js
chrisdickinson: Ah, this is a particularly gross bit of code where I'm overwriting the key in the original object with the result of the promise, so I wouldn't be able to use `Object.values` here (since I need the key to thunk out the attribute.)

services/registry/handlers/users.js
chrisdickinson: That would be great once this lands!

services/registry/handlers/maintainers.js
chrisdickinson: Haha, thanks! I got it from looking at Golang a while back, tbh.

services/common/boltzmann/request-handler.js
chrisdickinson: Ah yeah. `new URL` has the (uncomfortable in this case) prereq that you have to give it two args to prevent it from throwing if the first arg is an "incomplete" url – we _probably_ should move over to it, but as a follow up PR.

services/storage/handlers/namespaces.js
chrisdickinson: This _should_ appear a lot noisier in the logs now after adding dev-mode traces to error responses.

chrisdickinson:
OK, I'm calling it – this PR is ready to merge, I think. Brace yourself: I anticipate that it's going to break stuff and things will be a bit bumpy for a little while.

**Here's what I've done:**

- www.js has been split out of registry into `services/web`.
- Most of the registry code has moved back up into a new (private) service called `storage`.
    - The intent is that this will not be visible to the outside world.
- `web` and `registry` are responsible for trading external authentication (session cookies and authorization headers, respectively) for internal authentication (the surprisingly versatile `bearer: <username>` header.)
- `registry` caches authentication results in redis.
- membership (of users to namespaces) and maintainership (of packages by namespaces) URLs have been restructed **but the CLI has not caught up.** See **URL CHANGES** below.
- `docker-compose.yml` contains all the new services.
- Added `response.error.coded` which accepts a string (a la `namespace.members.bearer_unauthed`), designed to create a disjoint between `storage` layer error cases and `registry` layer error message display. (This means we can start pushing `i18n` concerns into the registry layer!)

**I've tested:**

- logging in both via services running locally and through docker
- publishing a new package
- testing `whoami`
- manually curling a few list endpoints – packages, maintainerships, memberships.

**I have not:**

- updated the documentation (for another PR!)
- added registry tests (but wow we need them! all of the existing tests – save invitations, which were deleted due to _massive upheaval_ – moved from registry to storage)
- audited the dependencies of each service. It's possible (likely!) that services are depending on things they aren't using.

## URL CHANGES

```
GET     /v1/users/user/:user/memberships[?pending=true]                list membership
POST    /v1/users/user/:user/memberships/invitations/:namespace@:host  accept invitation
DELETE  /v1/users/user/:user/memberships/invitations/:namespace@:host  decline invitatation

GET     /v1/namespaces/namespace/:namespace@:host/maintainerships[?pending=true]            list maintainerships
POST    /v1/namespaces/namespace/:namespace@:host/maintainerships/:namespace@:host/:package accept invite
DELETE  /v1/namespaces/namespace/:namespace@:host/maintainerships/:namespace@:host/:package decline invite

GET     /v1/namespaces/namespace/:namespace@:host/members[?pending=true] list members
POST    /v1/namespaces/namespace/:namespace@:host/members/:user          invite user
DELETE  /v1/namespaces/namespace/:namespace@:host/members/:user          remove member / rescind invite

GET     /v1/packages/package/:namespace@:host/maintainers[?pending=true] list maintainers
POST    /v1/packages/package/:namespace@:host/maintainers/:name@:host    invite maintainer
DELETE  /v1/packages/package/:namespace@:host/maintainers/:name@:host    remove maintainer / rescind invite
```

This eliminates all urls with the word `invite` in them, because we don't so much have invites as "pending" maintainerships and memberships; and the user/namespace is free to leave if they choose to do so. The CLI should be updated to reflect these changes! This can be done in a subsequent PR.

## why is it important to move all this cheese now

Moving this cheese:

- Gives people a service in which to build _all_ the website functionality in a connected way. (If we end up going with Next.js, we may need _yet another_ service that lives outside of the private network and uses `web` as an API gateway/backend-for-frontend. That's okay, though – we've already gone from 1 service to 3. 3 to 4 will be much easier.)
- Lets us separate expensive processes (like publishing and syncing!) into a worker process.

package.json
```diff
@@ -30,8 +30,8 @@
     "lint-fix": "prettier --write '**/*.js'",
     "lint-md": "markdownlint \"**/*.md\" -i \"**/node_modules/**\"",
     "lint-registry": "cd services/registry; npm run lint",
-    "postinstall": "for d in cli services/registry services/workers services/common/boltzmann; do cd $d; npm i; cd -; done",
+    "postinstall": "for d in cli services/{registry,workers,web,common/boltzmann}; do cd $d; npm i; cd -; done",
     "start": "docker-compose up",
-    "test": "for d in cli services/registry; do cd $d; npm t; cd -; done"
+    "test": "for d in cli services/{registry,workers,web,common/boltzmann}; do cd $d; npm t; cd -; done"
```
chrisdickinson: Use bash in anger.

services/common/boltzmann/client/NOTES
```diff
@@ -0,0 +1,100 @@
...
+    getPackage({namespace, host, name})
+    replacePackage({namespace, host, name, require_tfa})
+    deletePackage({namespace, host, name, ...data})
+    getPackageVersion({namespace, host, name, version})
+    createPackageVersion({namespace, host, name, version, request})
+    deletePackageVersion({namespace, host, name, version})
+    getObject({algo, digest}) // promise for a stream
+
+
+response.rawjson(str) // for enemies
```
chrisdickinson: Leaving this for now, will clean up afterward. Sectumsempra!

services/common/boltzmann/response.js
```diff
@@ -75,18 +77,32 @@ function error(err, status = 500, extraHeaders = {}) {
   if (typeof err === 'string') {
     err = { message: err, code: 'ENOTSUPPLIED' };
   }
+
+  if (isDev()) {
+    err.trace = new Error().stack;
+  }
```
chrisdickinson: If we're in dev mode, add a `trace` property to all outgoing errors. This makes it easier to track down errors when they happen: the registry client will log traces coming from `storage`.

services/common/boltzmann/response.js
```diff
@@ -75,18 +77,32 @@ function error(err, status = 500, extraHeaders = {}) {
...
+
   const r = new Response(JSON.stringify(err), { status, headers });
   return r;
 }
 
+error.coded = coded;
+function coded(code, ...args) {
+  return error(Object.assign(new Error(code), { code }), ...args);
```
chrisdickinson: For storage-layer use. Takes computer-friendly dotted, lowercase error codes.

services/common/boltzmann/response.js
```diff
@@ -75,18 +77,32 @@ function error(err, status = 500, extraHeaders = {}) {
...
+
 function authneeded(message, status = 401, extraHeaders = {}) {
   const headers = new Headers({
     'www-authenticate': 'bearer',
     'content-type': 'application/json',
     ...extraHeaders
   });
   if (typeof message === 'string') {
-    message = { error: message };
+    message = { message, code: 'authneeded' };
```
chrisdickinson: Standardize a bit on `message` (which is used in `response.error`), while we're at it.

services/registry/handlers/maintainers.js
```diff
@@ -16,88 +10,70 @@ module.exports = [
...
-    findInvitee(canWrite(remove))
-  ),
-  fork.post(
-    '/v1/packages/package/:namespace([^@]+)@:host/:name/invitation/:member',
-    packageExists(isNamespaceMember(accept))
-  ),
-  fork.del(
-    '/v1/packages/package/:namespace([^@]+)@:host/:name/invitation/:member',
-    packageExists(isNamespaceMember(decline))
+    authn.required(remove)
```
chrisdickinson: A theme emerges: registry endpoints can make authorization decisions on a large scale – "this will fail if you aren't authenticated" – but delegate granular authorization decisions to the storage layer ("can Alice write to Bob's package?")

services/registry/handlers/maintainers.js
```diff
@@ -16,88 +10,70 @@ module.exports = [
...
-    }
+    .then(xs => [null, xs], xs => [xs, null]);
+
+  if (err) {
+    const msg = {
+      'maintainer.invite.invitee_dne': `Unknown namespace: "${invitee}".`,
+      'maintainer.invite.package_dne': `Unknown package: "${invitee}".`,
+      'maintainer.invite.already_accepted': `Namespace "${invitee}" is already a member.`,
+      'maintainer.invite.already_declined': `Namespace "${invitee}" has declined this invite.`
+    }[err.code];
```
chrisdickinson: This is (more or less) what I'm thinking error enumeration should look like, per those many TODO comments.

services/registry/middleware/bearer-auth.js
```diff
@@ -1,26 +1,15 @@
...
+const crypto = require('crypto');
 
 module.exports = createBearerAuthMW;
 
-const idempotent = new Set(['GET', 'HEAD']);
-
-function createBearerAuthMW() {
+function createBearerAuthMW({
+  sessionTimeout = Number(process.env.SESSION_TIMEOUT) || 5 * 60
+} = {}) {
```
chrisdickinson: The registry layer _always_ attempts to trade the `Authorization` header for a `User` object when given. It'll try redis first, then fall back to asking the storage layer.

services/registry/middleware/bearer-auth.js
```diff
@@ -34,16 +23,34 @@ function createBearerAuthMW() {
...
-        const user = await Token.lookupUser(bearer);
-        if (user) {
-          context.user = user;
+        data = JSON.parse(data);
+        context.user = data;
+      } catch {}
+
+      if (data === null) {
+        const [err, result] = await context.storageApi
+          .getToken(bearer)
```
chrisdickinson: `getToken()` sends the token in a `{token}` header so sensitive data does not get logged. 🤫

services/registry/package.json
chrisdickinson: TODO another day, perhaps

services/storage/decorators/authn.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
```
chrisdickinson: The storage layer, however, only tries to load the User from the database if a handler _specifically requests it_ using one of these decorators.

services/storage/decorators/package-exists.js
```diff
@@ -0,0 +1,41 @@
...
+      namespace: 'namespace',
+      host: 'host',
+      name: 'name'
+    })(mapping);
+  }
+
+  const { namespace, host, name } = mapping;
+  return next => {
+    return async (context, params) => {
+      const { [namespace]: ns, [host]: h, [name]: pkg } = params;
```
chrisdickinson: ES2015: the idents are dynamic, and the points are all free style

services/storage/handlers/index.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
```
chrisdickinson: Most of the `storage/handlers` are ports of the corresponding registry files.

services/storage/handlers/users.js
```diff
@@ -0,0 +1,233 @@
...
+      modified: xs.modified,
+      active: xs.active,
+      accepted: xs.accepted
+    });
+  }
+
+  return response.json({
+    objects,
+    next: hasNext,
+    prev: hasPrev
```
chrisdickinson: A pagination scheme appears.

services/storage/middleware/internal-auth.js
```diff
@@ -0,0 +1,38 @@
...
+
+function createInternalAuthMW() {
+  return next => {
+    return async context => {
+      // Being lazy here, because we may not actually care
+      // to materialize the user on most requests. So, make
+      // `context.user` return a promise for the currently
+      // authenticated user (if any.)
+      //
+      // TODO: Someday use JWTs, maybe.
```
chrisdickinson: Maybe.

services/storage/models/authentication.js
```diff
@@ -1,60 +1,10 @@
...
-        redirect_uri: `${process.env.EXTERNAL_HOST}/www/login/providers/${
-          this.name
-        }/callback`,
-        state,
-        client_id: this.id
-      })
-    );
-  }
-}
-
```
chrisdickinson: Provider data lives in `web` now.

services/storage/models/authentication.js
```diff
@@ -96,6 +59,8 @@ module.exports = class Authentication {
   }
 };
 
+const User = require('./user');
```
chrisdickinson: Had to shuffle requires because of 🎶 circular imports :notes:

services/storage/models/token.js
```diff
@@ -34,6 +34,11 @@ module.exports = class Token {
     this.active = active;
   }
 
+  toJSON() {
+    const { user: _0, id: _1, user_id: _2, ...meta } = this;
```
chrisdickinson: Omitting properties by destructure+splat.

services/web/handlers/auth.js
chrisdickinson: Ideally this lands at `/settings/:username/tokens`, but I'm going to leave that for another PR.

services/web/handlers/index.js
chrisdickinson: It's a package registry on the move!

services/registry/handlers/index.js
```diff
@@ -4,23 +4,22 @@ module.exports = makeRouter;
...
     ...require('./maintainers'),
     ...require('./namespaces'),
-    ...require('./www'),
 
-    fork.get('/-/v1/login/poll/:session', auth.poll),
-    fork.post('/-/v1/login', auth.login),
-    fork.get('/v1/auth/whoami', isLoggedIn(whoami)),
+    fork.get('/-/v1/login/poll/:session', authn.anonymous(auth.poll)),
+    fork.post('/-/v1/login', authn.anonymous(auth.login)),
+    fork.get('/v1/auth/whoami', authn.required(whoami)),
```
ceejbot: I like the new approach toward naming the decorators. Gets rid of some unsightly camel case.

services/registry/handlers/index.js
```diff
@@ -37,11 +36,6 @@ async function version() {
   return response.json(data);
 }
 
-async function greeting() {
-  const objects = await User.objects.all().then();
-  return response.json({ objects });
-}
-
```
ceejbot: I *nearly* removed this in an earlier PR.

services/registry/handlers/maintainers.js
ceejbot: A somewhat familiar pattern: Hit redis for the session lookup, rewrite some headers, hand over the entire request. Making granular ACL decisions here would be wrong because of the back and forth. We're a gateway!

services/registry/handlers/maintainers.js
```diff
@@ -16,88 +10,70 @@ module.exports = [
...
-      active: true,
+  const [err, maintainers] = await context.storageApi
+    .listPackageMaintainers({
+      namespace,
+      host,
       name,
-      'namespace.active': true,
-      'namespace.name': namespace
+      bearer: context.user ? context.user.name : null,
+      page: Number(context.url.query.page) || 0,
```
ceejbot: pagination rears its ugly head for the first time in this code base

services/registry/handlers/maintainers.js
```diff
@@ -16,88 +10,70 @@ module.exports = [
...
+      'maintainer.invite.invitee_dne': `Unknown namespace: "${invitee}".`,
+      'maintainer.invite.package_dne': `Unknown package: "${invitee}".`,
+      'maintainer.invite.already_accepted': `Namespace "${invitee}" is already a member.`,
+      'maintainer.invite.already_declined': `Namespace "${invitee}" has declined this invite.`
+    }[err.code];
+    return response.error(
+      msg ||
+        `Caught error inviting "${invitee}" to ${namespace}@${host}/${name}`,
+      err.code
+    );
```
ceejbot: This ends up being tidy & a reasonable pattern to repeat, I think.

services/registry/handlers/maintainers.js
```diff
@@ -109,35 +85,32 @@ async function invite(context, { namespace, host, name, invitee }) {
...
+  const [err, invite] = await context.storageApi
+    .removePackageMaintainer({
+      namespace,
+      host,
+      name,
+      bearer: context.user.name,
+      to: invitee
     })
-    .then();
+    .then(xs => [null, xs], xs => [xs, null]);
```
ceejbot: I feel your `xs` habit here makes this harder to read than it would be if you named the args invitation & err.

services/registry/handlers/maintainers.js
```diff
@@ -109,35 +85,32 @@ async function invite(context, { namespace, host, name, invitee }) {
...
+      'maintainer.invite.invitee_not_maintainer': `${invitee} was not a maintainer of ${namespace}@${host}/${name}.`,
+      'maintainer.invite.package_dne': `Unknown package: "${invitee}".`,
+      'maintainer.invite.already_accepted': `Namespace "${invitee}" is already a member.`,
+      'maintainer.invite.already_declined': `Namespace "${invitee}" has declined this invite.`
+    }[err.code];
+
+    return response.error(
+      msg ||
+        `Caught error inviting "${invitee}" to ${namespace}@${host}/${name}`,
+      err.code
```
ceejbot: Yeah, on second appearance I continue to like it.

services/registry/handlers/packages.js
```diff
@@ -246,261 +115,73 @@ async function packageDelete(context, { host, namespace, name }) {
...
-    .catch(PackageVersion.objects.NotFound, () => null);
+    .then(xs => [null, xs], xs => [xs, null]);
 
-  if (!v) {
-    return response.error(
-      `Could not find "${namespace}@${host}/${name} at ${version}"`,
-      404
-    );
+  if (err) {
+    // TODO: enumerate errors
```
ceejbot: everything is always 80% error handling at least

services/registry/handlers/packages.js
```diff
@@ -246,261 +115,73 @@ async function packageDelete(context, { host, namespace, name }) {
...
-    Object.keys(formdata.dependencies).length +
-      Object.keys(formdata.optionalDependencies).length +
-      Object.keys(formdata.devDependencies).length +
-      Object.keys(formdata.peerDependencies).length >
-    MAX_DEPENDENCIES
-  ) {
-    return response.error(`Exceeded maximum number of dependencies.`, 400);
-  }
+async function getObject(context, { algo, '*': digest }) {
+  const [err, stream] = await context.storageApi
```
ceejbot: While we did already have this pluggable in `context.storage.strategy`, the implementation seems even more hidden now. This could be going to an entirely separate service than everything else if scale someday demands that.

services/registry/middleware/bearer-auth.js
ceejbot: Ideally this ends up being a redis hit almost every time, and only rarely does satisfying the full request require two full roundtrips + db hits.

services/registry/middleware/bearer-auth.js
```diff
@@ -34,16 +23,34 @@ function createBearerAuthMW() {
         );
       }
 
+      // getting access to the redis doesn't get you the tokens.
```
ceejbot: good

services/registry/middleware/bearer-auth.js
```diff
@@ -34,16 +23,34 @@ function createBearerAuthMW() {
...
+        const [err, result] = await context.storageApi
+          .getToken(bearer)
+          .then(xs => [null, xs], xs => [xs, null]);
+
+        if (!err) {
+          await context.redis.setexAsync(
+            key,
+            sessionTimeout,
+            JSON.stringify(result.user)
+          );
```
ceejbot: Probably don't need to wait for this to happen, and probably don't even want to crash on out of the request if it fails. Should fix later.

services/registry/package.json
ceejbot: writing up issue

services/storage/decorators/authn.js
ceejbot: As before, I like the new decorating naming.

services/storage/handlers/users.js
ceejbot: which we'll eventually need to punch through to the gateway. unless it is already and I failed to notice

services/storage/middleware/internal-auth.js
ceejbot: Grumpy sound.

services/storage/models/user.js
```diff
@@ -30,6 +30,11 @@ module.exports = class User {
     this.active = active;
   }
 
+  toJSON() {
+    const { tfa_secret: _0, backup_codes: _1, id: _2, ...meta } = this;
```
ceejbot: Good choices to omit.

services/storage/models/token.js
ceejbot: Fiendish!

services/web/handlers/auth.js
```diff
@@ -359,24 +371,23 @@ async function handleTokenAction(context) {
...
-      .update({ active: false });
+    const { count } = await context.storageApi.deleteToken({
+      for: user,
+      valueHashes: [token]
+    });
 
-    context.session.set('banner', 'Successfully deleted 1 token.');
+    context.session.set(
+      'banner',
+      `Successfully deleted ${count} token${count === 1 ? '' : 's'}.`
```
ceejbot: this is what an acceptable ternary looks like
memorize this chris
don't nest them
no stop don't

services/web/server.js
zacanger: I made a ticket for that [here](https://github.com/entropic-dev/entropic/issues/213) so it doesn't get lost

services/storage/handlers/packages.js
zacanger: Added [a ticket](https://github.com/entropic-dev/entropic/issues/214) so this doesn't get lost.

PR #138, opened 2019-06-04
Added help menus, prevents ugly crash

wbrickner:
Description of changes:

- Added help menus,
- Prevented crash when unknown command was entered (for instance, `ds install`, as I'm sure many VCPM people will constantly enter at first),
- Minor additions and corrections to printed text

P.S. Great talk!

### DIFF
--- a/cli/lib/commands/help/--version.txt
+++ b/cli/lib/commands/help/--version.txt
@@ -1,0 +1,4 @@
+Usage: ds --version
+
+Prints the version of ds you're using
+
--- a/cli/lib/commands/help/login.txt
+++ b/cli/lib/commands/help/login.txt
@@ -1,0 +1,6 @@
+Usage: ds login [--registry url]
+
+1) ds opens the default web browser, 
+2) you authenticate, 
+3) ds stores the resulting authentication token internally.
+
--- a/cli/lib/commands/help/whoami.txt
+++ b/cli/lib/commands/help/whoami.txt
@@ -1,0 +1,4 @@
+Usage: ds whoami
+
+Contacts your registry and figures out which user you're logged in as.
+
--- a/cli/lib/main.js
+++ b/cli/lib/main.js
@@ -9,12 +9,17 @@ const minimist = require('minimist');
 const { load } = require('./config');
 
 async function main(argv) {
-  if (!argv[0]) {
-    argv[0] = 'help';
-  }
+  if (!argv[0]) { argv[0] = 'help'; }
 
   try {
-    const cmd = require(`./commands/${argv[0]}`);
+    let cmd 
+    try { cmd = require(`./commands/${argv[0]}`); }
+    catch (e) {
+      // this command isn't feeling well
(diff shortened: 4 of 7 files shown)


cli/lib/main.js
```diff
@@ -9,12 +9,17 @@ const minimist = require('minimist');
...
-  }
+  if (!argv[0]) { argv[0] = 'help'; }
 
   try {
-    const cmd = require(`./commands/${argv[0]}`);
+    let cmd 
+    try { cmd = require(`./commands/${argv[0]}`); }
+    catch (e) {
+      // this command isn't feeling well
+      return console.error(`An error ocurred when trying to "${argv[0]}". \n` +
```
varjmes: Was about to open a PR for something similar :)

cli/lib/commands/help/--version.txt
```diff
@@ -0,0 +1,4 @@
+Usage: ds --version
```
varjmes: Could we throw in an alias to `-v` somehow? I keep on typing `ds -v` and then making a sad face.

cli/lib/commands/help/login.txt
```diff
@@ -0,0 +1,6 @@
+Usage: ds login [--registry url]
+
+1) ds opens the default web browser, 
+2) you authenticate, 
+3) ds stores the resulting authentication token internally.
```
varjmes:
```suggestion
3) ds stores the resulting authentication token within your .entropicrc file
```

cli/lib/commands/help/whoami.txt
```diff
@@ -0,0 +1,4 @@
+Usage: ds whoami
```
varjmes:
```suggestion
Usage: ds whoami [--registry url]
```

cli/lib/main.js
```diff
@@ -9,12 +9,17 @@ const minimist = require('minimist');
 const { load } = require('./config');
 
 async function main(argv) {
-  if (!argv[0]) {
-    argv[0] = 'help';
-  }
+  argv[0] = argv[0] || 'help';
```
zacanger: Is mutating argv safe, or would it make sense to do something like `const args = [...argv]; args[0] = args[0] || 'help'` here?

wbrickner: Can you expand on what you mean by safety?  What are some dangerous side effects you expect could occur?

This was the effect of the original code from `master`.

zacanger: My main concern is about mutating actual `process.argv`, which could cause side effects if, for example, someone is executing `ds` from another script. But I just read the rest of the file and saw `main(process.argv.slice(2))`. Sorry I missed that before!

wbrickner: Great, so for posterity the `argv` we're dealing with is a different object, so you won't get those types of unexpected side effects.

wbrickner:
Wondering what the timeline looks like to merge this PR?

PR #148, opened 2019-06-05
[CLI] - validate commands and exit if invalid

olingern:
### Issue
The CLI should not crash when given an invalid command. Found this when playing around with the registry.

DeepinScreenshot_select-area_20190605102813

### Proposal
Validate commands before attempting to require the module

### Tests
Adding a test would be dependent upon #118. I can create an empty file with a `#TODO` or open an issue.


olingern:
Ah, I just saw #138

PR #134, opened 2019-06-04
chore: Add Markdownlint

nschonni:
Add the intial setup, but ignore all current issues

PR #149, opened 2019-06-05
[CI] - Add lint checking

olingern:
### Description

I noticed that when running `npm run lint-fix` that some files were committed that were not fixed. I also sometimes forget to lint before pushing. I would expect the CI to fail if `npm lint-fix` still has things it autofixes as 40797f27698fc31e86cbe92fc597569a805f32e3 does

This PR adds a repo-wide lint command, `lint-check`, using [prettier check](https://prettier.io/docs/en/cli.html#check).

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
   "repository": "https://github.com/entropic-dev/entropic",
   "scripts": {
     "lint": "run-s -s lint-registry lint-cli lint-md",
+    "lint-check": "prettier --check '**/*.js'",
     "lint-cli": "cd cli; npm run lint",
     "lint-fix": "prettier --write '**/*.js'",
     "lint-md": "markdownlint \"**/*.md\" -i \"**/node_modules/**\"",
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,19 @@
 version: 2
 jobs:
+  lint:
+    docker:
+      - image: circleci/node:latest
+    working_directory: ~/repo
+    steps:
+      - checkout:
+          path: ~/repo
+      - restore_cache:
+          keys:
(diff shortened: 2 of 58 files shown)


package.json
```diff
@@ -28,6 +28,7 @@
   },
   "scripts": {
     "lint": "npm run lint-registry && npm run lint-cli",
+    "lint-check": "prettier --check '**/*.js'",
```
nschonni: Should this get chained into the `lint` target?

olingern: Hmm, that seems like it could be helpful. If I did, I think I would embed it into `lint-cli` and `lint-registry`, but I saw this as project-wide linting specifically for the CI.

I'm not really sure what's ergonomically preferable. Maybe keeping `lint-check` in the root for project-wide linting in the CI, and then tacking `prettier --check` onto both `lint-cli` and `lint-registry`?

nschonni: Not sure, I submitted #134 and figured that if this added the general `lint` it would cover both cases :smile:
I'm not sure if the preference in the project is for subpackage targets or not

olingern:
Ok, rebased for post #135 merge. For the lint command I also _only_ install prettier because we don't need to reinstall all dependencies at this step.

olingern:
Rebased. This is going to fail until someone makes a lint commit.

olingern:
K, rebased again. and added the `lint-fix` commit in c33be2878af35ec392cfec09a3e80564c50478b9

olingern:
@chrisdickinson @ceejbot sorry to ping :bowing_man: , but I think this one may get painful for other people after more code is merged

nschonni:
@all-contributors add @olingern for infra

Trying this out because I landed something upstream so the contributors table should be pretty-printed now

PR #165, opened 2019-06-05
chore: Allow "a" in markdown

nschonni:
Used for the alternate TOC in Contributing.md


nschonni:
Alternately, the one instance could be disabled inline

PR #174, opened 2019-06-05
chore: Add docs CI setup

nschonni:


nschonni:
Failure is expected till #173 lands

.circleci/config.yml
```diff
@@ -44,9 +44,32 @@ jobs:
 
       - run: npm test
 
+  docs:
```
zacanger: Might be worth using an "anchor" for some of these repeated blocks. https://learnxinyminutes.com/docs/yaml/ I would point to the yaml spec but it's like 200 pages

.circleci/config.yml
```diff
@@ -44,9 +44,32 @@ jobs:
...
+    docker:
+      - image: circleci/node:latest
+      - image: circleci/postgres:latest
```
zacanger: Probably no need for postgres heree

nschonni: Thanks, bad copy/paste

nschonni:
@chrisdickinson no problem, I rebased and added an additional suppression for details/summary since one was added since the original PR

chrisdickinson:
Merged in 44ad820! Thanks!

PR #176, opened 2019-06-06
Switch var to let and const

zacanger:
Just a little thing that was bugging me. I noticed everywhere else uses `const` and `let`, but a couple places in the CLI were using `var`.

### DIFF
--- a/cli/lib/commands/whoami.js
+++ b/cli/lib/commands/whoami.js
@@ -20,7 +20,7 @@ async function whoami(opts) {
     }
   });
 
-  var body = null;
+  let body = null;
   try {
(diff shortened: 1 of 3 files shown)


nschonni:
Maybe enabling this ESLint rule would make sense then? https://eslint.org/docs/rules/no-var#require-let-or-const-instead-of-var-no-var
`eclint . --fix` with that open on, cleans up a few others too

zacanger:
That seems like a good idea! I haven't wanted to touch the eslint config because I have lots  of very strong opinions that I don't want to try to force on other people 😳

PR #178, opened 2019-06-06
[WIP] - ds init command

olingern:
Initial prototype for the `ds init` command.

Currently, this command initializes a `Package.toml`

### Needs these to merge before finishing
- #189 --> `Package.toml` 
  Need to determine final form of `toml` file

- ~#118 --> Ava~ Replaced by #225
- ~#111 --> Chalk~ Replaced by #225

### Questions
- What is a valid `name`? 
    - Will it always consist of `name@registry-domain.dev/package_name`?  

### Currently

ezgif com-optimize


varjmes:
How about `init`? In no way trying to copy other CLI tools, but I feel it fits better than create. Create is often used as the first arg of a two arg command, e.g. `ds create entity` `ds create snargleflargle`.

olingern:
@varjmes `init` does feel more familiar, but I was largely going off of the CLI, ["Commands that should exist"](https://github.com/entropic-dev/entropic/tree/master/cli#commands). 

Totally open to whatever people generally feel best about typing.

superhawk610:
+1 for `init` over `create`

chrisdickinson:
Ooh, there's some distinction here. In entropic, you can create a package on the registry _without publishing any versions to it_. `publish` automatically creates the package if it detects it missing. You can see this [here](https://github.com/entropic-dev/entropic/blob/master/cli/lib/commands/publish.js#L94-L124). It exists so you can register a package without having the code ready / set up initial security settings like 2FA ahead of publishing. (It's sort of like a sanctioned typosquat, but since everything is namespaced & federated it's not quite so big of a deal.)

`init`, on the other hand should purely initialize a directory for work.

olingern:
@chrisdickinson Ah, I see. So I've created an `init` command here. Just so I understand, `ds create` would be use to park your virtual car in space you'd like, which would then allow you to do some admin things like you mention (2FA, etc.). 

I'll update and rename this command, but maybe I can pull some ideas from #189 as to how a `Package.toml` should be initialized.

olingern:
@ceejbot gotcha! I'd be happy to contribute to a "terminology" doc so we don't lose all this useful info in these issues.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,203 @@
...
+      rl,
+      validateVersion
+    );
+    answers['entry'] = await askQuestion(QUESTIONS.ENTRY, rl);
+    answers['type'] = await askQuestion(QUESTIONS.TYPE, rl);
+    answers['license'] = await askQuestion(QUESTIONS.LICENSE, rl);
+    answers['description'] = await askQuestion(QUESTIONS.DESCRIPTION, rl);
+    answers['homepage'] = await askQuestion(QUESTIONS.HOMEPAGE, rl);
+    answers['author'] = await askQuestion(QUESTIONS.AUTHOR, rl);
+    answers['repository'] = await askQuestion(QUESTIONS.REPOSITORY, rl);
```
olingern: I can be more clever and provide some sort of abstraction, but I'm not sure at what cost it will come in terms of readability.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+      );
+
+      if (VALID_YES_NO[proceed] === 0) {
+        console.log('Exiting.');
+        rl.close();
+        return 0;
+      }
+    }
+
+    let answers = [];
```
zacanger: Should this be an object?

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+
+    rl.close();
+
+    await writeFile(tomlLocation(), createToml(answers));
+
+    if (answers['type'] === 'module') {
+      await writeFile(process.cwd(), createPackageJson());
+    }
+  } catch (e) {
+    console.error('There was an error creating your Package.toml');
```
viktor-ku: Avoid `try / catch` that covers too much of code. Operation specific errors are generally a lot easier to digest and work with.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+/**
+ * Asks user a question until a valid answer is given and applies any
+ * desired transformations to the value returned if provided
+ */
+async function askQuestion(question, rl, validator, transform = undefined) {
+  let invalid = true;
+  let ans = undefined;
+  const { MESSAGE, REQUIRED } = question;
+
+  while (invalid) {
```
viktor-ku: If you were to go with this approach `while(true)` could be used. `invalid` variable could be thrown away. `invalid = false;` could be then replaced with just `break`

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+    if (fileExists(tomlLocation())) {
+      // Ask if we should proceed since the Package.toml exits
+      const proceed = await askQuestion(
+        QUESTIONS.TOML_EXISTS,
+        rl,
+        validateYesNo,
+        lowercase
+      );
+
+      if (VALID_YES_NO[proceed] === 0) {
```
viktor-ku: `I_AM_CONST = 42` uppercase words are usually used to define string or number constants. This line is super confusing because I have no idea what to expect from it and why are you trying to access the string.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+ */
+function validateName(name) {
+  // TODO: Uese Regex for "name@registry-domain.dev/package_name"?
+  return name && name.length > 0;
+}
+
+/**
+ * Validates the version provided by the user
+ */
+function validateVersion(version) {
```
viktor-ku: One line wrappers are generally not needed/useful.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+    return false;
+  }
+
+  return VALID_YES_NO[ans] !== undefined;
+}
+
+/**
+ * Validates the name provided by the user
+ */
+function validateName(name) {
```
viktor-ku: Rename function name to reflect the type of name you are validating here.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+    MESSAGE: 'Repository: ',
+    REQUIRED: false
+  }
+};
+
+/**
+ * Returns string for the path to write Package.toml
+ */
+function tomlLocation() {
+  return `${process.cwd()}/Package.toml`;
```
viktor-ku: 1. Use `path.join`. It will be more consistent (check out the rest of the app)
2. Make this a simple variable. Function is definitely an overkill for this.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+};
+
+const VALID_YES_NO = {
+  y: 1,
+  yes: 1,
+  n: 0,
+  no: 0
+};
+
+const QUESTIONS = {
```
viktor-ku: It's a plain object. No need to uppercase everything. Please use `camelCase` and lowercase. Will be much more readable.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+  writeFile
+} = require('../utils');
+
+const REJECTION_MSGS = {
+  REQUIRED: 'This is required',
+  validateName: "Sorry, that's not a valid name",
+  validateVersion: "Sorry, that's not a valid semver version"
+};
+
+const VALID_YES_NO = {
```
viktor-ku: I guess `1` is `true` and `0` is `false` like in good old `C` times? Well booleans are completely fine :+1:

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+  VERSION: {
+    MESSAGE: 'Version: ',
+    REQUIRED: true
+  },
+  ENTRY: {
+    MESSAGE: 'Entry: ',
+    REQUIRED: false
+  },
+  TYPE: {
+    MESSAGE: 'Type :',
```
viktor-ku: I've noticed an inconsistency here - sometimes you would have space before `:` and sometimes you don't

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+ * Validates the version provided by the user
+ */
+function validateVersion(version) {
+  return isValidSemver(version);
+}
+
+/**
+ * Used to provide consistent casing for hash lookups
+ */
+function lowercase(str) {
```
viktor-ku: Not needed at all. Down below you know that you need to lowercaes everything so just use `lowercase` in its normal form there.

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+}
+
+/**
+ * Minimal prompt for asking users for input
+ */
+function ask(question, rl) {
+  return new Promise((resolve, reject) => {
+    rl.question(question, answer => resolve(answer));
+  }).catch(e => {
+    console.error(e);
```
viktor-ku: Is this thing able to throw an error? What do you need to do for it to throw one?

cli/lib/commands/init.js
olingern: Yes! I have a Japanese keyboard and sometimes make this mistake :slightly_frowning_face:

cli/lib/commands/init.js
olingern: There lies an edge case somewhere where this promise might get rejected. If an error does happen, a user should be able to open an issue with the log rather than just saying, "it doesn't work"

cli/lib/commands/init.js
olingern: It's used for flexible augmentation of the answer. As the RFC pans out, I anticipate more transformations to be added. If not, I'll move when it's finalized.

cli/lib/commands/init.js
olingern: updated

cli/lib/commands/init.js
viktor-ku: Right. But the error will be thrown with a message and a stack anyway. I think you don't need this catch here.

cli/lib/commands/init.js
olingern: I went back and forth because uppercasing tells me these are constants without having to wonder or look around. I've lowercased them since `const` will enforce it.

cli/lib/utils.js
```diff
@@ -0,0 +1,72 @@
...
+    return false;
+  }
+
+  const semverRegex = /^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(-(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(\.(0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*)?(\+[0-9a-zA-Z-]+(\.[0-9a-zA-Z-]+)*)?$/;
+  return input.match(semverRegex);
+}
+
+function optionalEntries(answers) {
+  return Object.keys(answers)
+    .filter(k => answers[k].length > 0 && k !== 'name' && k !== 'version')
```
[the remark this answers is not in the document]
olingern: `Object`

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
+'use strict';
+
+module.exports = create;
```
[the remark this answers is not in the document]
olingern: updated

cli/lib/commands/init.js
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  return ans;
+}
+
+/**
+ * Exported function used as `ds create`
+ */
+async function create(opts) {
+  await writeFile(`${process.cwd()}/package.json`, createPackageJson());
```
superhawk610: For consistency, I'd either get rid of `tomlLocation` or create an additional helper for `jsonLocation`.

PR #183, opened 2019-06-06
chore(cli): refactors the fetch

pranshuchittora:
Refactored the fetch.js

### DIFF
--- a/cli/lib/fetch.js
+++ b/cli/lib/fetch.js
@@ -19,11 +19,9 @@ function fetchWithAgent(resource, init) {
     if (!init) init = {};
 
     const url = new URL(resource);
-    if (url.protocol == 'https:') {
-        init.agent = httpsAgent;
-    } else {
-        init.agent = httpAgent;
-    }
+    url.protocol == 'https:'
+      ? (init.agent = httpsAgent)
+      : (init.agent = httpAgent);
 
     return fetch(resource, init);
 }


cli/lib/fetch.js
```diff
@@ -19,11 +19,9 @@ function fetchWithAgent(resource, init) {
     if (!init) init = {};
 
     const url = new URL(resource);
-    if (url.protocol == 'https:') {
-        init.agent = httpsAgent;
-    } else {
-        init.agent = httpAgent;
-    }
+    url.protocol == 'https:'
```
uchibeke: Wondering why we don't do:

```
init.agent = url.protocol == 'https:' ? httpsAgent : httpAgent
```

pranshuchittora: :+1:

cli/lib/fetch.js
```diff
@@ -9,22 +9,19 @@ const fetch = require('node-fetch');
...
+  hints: dns.ADDRCONFIG | dns.V4MAPPED
+});
 const httpsAgent = new https.Agent({
-    family: 6, hints: dns.ADDRCONFIG | dns.V4MAPPED
-})
+  family: 6,
+  hints: dns.ADDRCONFIG | dns.V4MAPPED
+});
 
 function fetchWithAgent(resource, init) {
```
zacanger: This line could be `function fetchWithAgent(resource, init = {}) {` rather than including the check below.

cli/lib/fetch.js
```diff
@@ -9,22 +9,19 @@ const fetch = require('node-fetch');
 const https = require('https');
 
 const httpAgent = new http.Agent({
-    family: 6, hints: dns.ADDRCONFIG | dns.V4MAPPED
-})
+  family: 6,
```
olingern: Node docs are really sparse on this, but `4` and `6` are actually used for the IP address family (IPv4 and IPv6). 

It would be nice if there were a constant for this:

```javascript
const IP_FAMILIES: {
  V4: 4,
  V6: 6
}
```
without being explicit, it looks like a ["magic number"](https://en.wikipedia.org/wiki/Magic_number_(programming)) to me.

Issue #104, opened 2019-06-01
KNOWN ISSUE: `ds login` fails after install (fix: run ds using Node >= 12)


chrisdickinson:
Pinning this issue (first reported by @flxwu & underlying issue identified by @itsananderson – thanks!)

If you're running into issues logging in after installing `ds`, check your Node version. At the time of writing, `ds` only supports Node 12+.

flxwu:
What about implementing a polyfill for Node < 12?

moogacs:
I want to work on it ?
Suppose the error happens actually because the `dir` is not exists , under any way should be created if not exists ?

PR #185, opened 2019-06-06
fix(#104): fix fs undefined in node < 12 

moogacs:


cli/lib/errors.js
```diff
@@ -2,5 +2,6 @@
 
 module.exports = {
   CouldNotReadConfigFile: class extends Error {},
-  CouldNotParseConfigToml: class extends Error {}
```
viktor-ku: I am using this error message in this PR https://github.com/entropic-dev/entropic/pull/186

varjmes: The error isn't removed. See the code below this line

moogacs:
I brought it back ?

cli/lib/config.js
```diff
@@ -2,35 +2,37 @@
...
-
-  var parsed = null;
-  try {
-    parsed = toml.parse(content);
-  } catch (e) {
-    throw new errors.CouldNotParseConfigToml(filename, e);
-  }
+  var content = '';
+  var parsed = '';
+  await fs.readFile(filename, 'utf8', (e, data) => {
```
olingern: I think using the fs.promises ( `require('fs').promises` ) is cleaner for this. Also, here you're `await`ing a callback.

If you need to do some specific handling for node versions, I would use [`process.version`](https://nodejs.org/api/process.html#process_process_version) and handle accordingly. 

Though, I know the docs say that Node 12 is required, so it's worth asking @chrisdickinson if this is needed.

moogacs:
@chrisdickinson , We need your opinion 😊

cli/lib/config.js
viktor-ku: Why would you have a callback and `await` used together?

cli/lib/config.js
```diff
@@ -2,35 +2,37 @@
...
-    parsed = toml.parse(content);
-  } catch (e) {
-    throw new errors.CouldNotParseConfigToml(filename, e);
-  }
+  var content = '';
+  var parsed = '';
+  await fs.readFile(filename, 'utf8', (e, data) => {    
+    if (e.code === 'ENOENT') {      
+      if (!fs.existsSync(filename)){
+        fs.mkdirSync(filename);
```
viktor-ku: Why do you need to make directory?

moogacs: Suppose to be the dir for .toml files 🤔

cli/lib/config.js
moogacs: That's true, doesn't make sense , shall i update it or will go with #186

viktor-ku: I'd wait for #186

cli/lib/config.js
viktor-ku: I see. Good intention, but as of now I see only one file in the home directory to be created + one in the root of the project.

moogacs:
@viktor-ku  well, after reverting and adding some new stuff , it seems got messy !

moogacs:
@zacanger 👍 that's better

moogacs:
@zacanger that's good 👍

moogacs:
@ceejbot already added with #216   , the only different thing now is #211 , so I don't if shall I revert all the previous commits then push again to resolve those conflicts , 
#104 not decided yet

Issue #162, opened 2019-06-05
Design: why a TOML file? (Incompatibility with existing ecosystem)


wbrickner:
Hello,

I've noticed that Entropic packages are described by TOML files (e.g. `Package.toml`), rather than the existing standard from VCPM (`package.json`).

One thing that is hard to ignore that I love about `package.json`, is that I can do run-time introspection on the current project (or modules I depend on).

For instance, it's a single line to get an object representation of a `package.json`:

```javascript
const self = require("package.json")
```

Whereas I imagine it's a bit more painful if I'm dealing with `Package.toml`.

I'm not sure if this is a common pattern, but it's something to consider, because I imagine there are a host of dev tools which expect `package.json`.

Why was the decision made to use TOML over JSON?  It seems to demand more background knowledge of new users (especially those coming from VCPM).

Is there some performance reason?  I know that Entropic is not only a node package manager, and so maybe these concerns are inappropriate for a package manager designed to handle packages for many languages and environments. Although, with the ubiquity of JSON, it seems odd to choose TOML.

On the other hand, I suppose it would be somewhat valuable to visually distinguish an Entropic package from another package, but maybe this is the wrong reason to make such a decision.

Can anyone shed any light on the advantages of TOML that I'm missing?

chrisdickinson:
Previously: https://github.com/entropic-dev/entropic/issues/106#issuecomment-497982913

> We made the decision to switch to `Package.toml` for a couple of reasons (that really need to be documented in `docs/README.md`):
> 
>     * Separates the computer-oriented "manifest" info from the human oriented "shipping" info. That is: a human writes a `Package.toml`, which the CLI interprets in order to create a derived `manifest` on the server.
> 
>     * Only reading `Package.toml` lets package authors publish a single package to both entropic and legacy registries, and lets them "opt in" to publishing to an entropic instance gradually.
> 
>     * Entropic package specifications (`chris@localhost:3000/example`) are invalid in both the `name` and `dependencies` key position for legacy packages. Existing CLIs will fail if they see them there, so it forces authors to make an up-front, final decision about which registry they're publishing to, or else swap files on publish.
> 
> 
> For these reasons it's pretty unlikely that `ds` will grow support for reading `package.json`. That said, I think we'll probably have to emit package.json files when materializing files to disk because Node relies on them to determine, e.g., where the `main` module lives on disk. (Until Tink or import maps land.)

wbrickner:
[...]

Issue #147, opened 2019-06-05
Package.toml versus package.toml (casing)


ravinggenius:
Is it possible to use the all-lowercase `package.toml`, or does the "p" need to be capitalized? The README spells it `Package.toml` in two places, even though it isn't at the beginning of the sentence. Was that intentional? Personally I'd prefer to keep it lowercase.

varjmes:
I tried with `package.toml` and was successful. Although here https://github.com/entropic-dev/entropic/blob/f6e1ccf304a7b2d535b092ad86374ffccc243e5f/cli/lib/load-package-toml.js#L12 it asks for `Package.toml` in the code, I am using OSX which is case insensitive, like Windows.

viktor-ku:
I don't get capitalized `P` as well

zacanger:
For consistency with the legacy node package manager, and not appearing higher up in `ls` results than things like README and CHANGES, I would also love it if we went with `package.toml`, lowercase.

toddself:
`P[tab]` = `Package.toml`
`p[tab]` = `package.json`

tab completion for the win!

zacanger:
Except on Windows which doesn't know what case is, Macs which know what case is but just don't care about it, and anyone who's got `set completion-ignore-case on` in their inputrc.

toddself:
yeah sorry that wasn't super helpful as a comment :)

zacanger:
No worries, it's totally valid if you're on Linux or BSD (we should all just be on BSD, we'd be so much more hipster)

ceejbot:
I believe Chris picked capital-P Package.toml because Rust uses capital-C `Cargo.toml`, and of course we're iterating on `package.json`.

If I let myself bikeshed this, my thoughts are:

- We might want to name it `entropic.toml`, following the Rust model. "Package" feels too generic.
- All lower case better because [insert lots of reasons here].

But I think it's not the end of the world if it stays `Package.toml`.

chrisdickinson:
> `P[tab]` = `Package.toml`
> `p[tab]` = `package.json`
> 
> tab completion for the win!

> I believe Chris picked capital-P Package.toml because Rust uses capital-C `Cargo.toml`, and of course we're iterating on `package.json`.

Both of these factored in. The potential for bugs when folks collaborate on entropic packages between case-sensitive and case-insensitive filesystems makes me wince a bit, though.

[...]

Issue #106, opened 2019-06-01
package.json support


DanielMazurkiewicz:
Supporting json based description of package seems pretty important, even support for incompatible version of package.json would be a plus.

And my personal preference would be also adding an option to load package description as a js object (without necessity of placing quotes) and with possibility to define scripts not only as a shell commands, but also as functions to which command line arguments and environment variables are passed.

```
{
  name: "chris@entropic.dev/ds",
  version: "0.0.0-beta",

  scripts: {
    test: "node test.js" // classic shell script
    compile: function(args, env, shell) {
      shell.run("ts " + args[0]);
    }
  },

  dependencies: {
      "@iarna/toml": "^2.2.3",
      "legacy@entropic.dev/figgy-pudding": "^3.5.1"
  }
}
```

Your solution could support all these 3 formats (toml > json > js like)

chrisdickinson:
We made the decision to switch to `Package.toml` for a couple of reasons (that really need to be documented in `docs/README.md`):

- Separates the computer-oriented "manifest" info from the human oriented "shipping" info. That is: a human writes a `Package.toml`, which the CLI interprets in order to create a derived `manifest` on the server.
- Only reading `Package.toml` lets package authors publish a single package to both entropic and legacy registries, and lets them "opt in" to publishing to an entropic instance gradually.
- Entropic package specifications (`chris@localhost:3000/example`) are invalid in both the `name` and `dependencies` key position for legacy packages. Existing CLIs will fail if they see them there, so it forces authors to make an up-front, final decision about which registry they're publishing to, or else swap files on publish.

For these reasons it's pretty unlikely that `ds` will grow support for reading `package.json`. That said, I think we'll probably have to emit package.json files when materializing files to disk because Node relies on them to determine, e.g., where the `main` module lives on disk. (Until Tink or import maps land.)

toddself:
@DanielMazurkiewicz i just wrote this tool though: https://github.com/toddself/package2toml that will read in your package.json and generate a Package.toml file from it!

ceejbot:
OH THAT IS SO NICE.

ForsakenHarmony:
have you considered yaml?

Or at least giving the top level thing a name, because toml is kinda fucked

toddself:
@ForsakenHarmony personally i've never enjoyed YAML as a specification format.  [yaml: probably not so great after all](https://arp242.net/yaml-config.html) covers a lot of the issues I have with yaml.

[...]

PR #189, opened 2019-06-06
Add RFC for manifest toml format

toddself:
There are a few things that we likely can add to this, but this covers
the format as it is along with an "entrypoint" specification which we
need currently!

### DIFF
--- a/docs/RFCs/001-manifest-toml.md
+++ b/docs/RFCs/001-manifest-toml.md
@@ -67,7 +67,7 @@ These keys are **NOT REQUIRED** be present when interacting with the CLI locally
 The manifest **MAY** contain the following fields:
 
 - `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `[TK].toml` file. e.g `"./src/index.js"`.
-- `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), and will signal to the node interpreter to run this code an ES6 Module.
+- `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field). The use of this field will explain to the interpreter how it should handle`.js` files.
 - `"license"` - a valid [SPDX license](https://spdx.org/licenses/) for your code. If a license is not provided then the code is considered to be `UNLICENSED`.
 - `"description"` - a short description of your package
 - `"homepage"` - a URL for your project's homepage
--- a/docs/RFCs/001-manifest-toml.md
+++ b/docs/RFCs/001-manifest-toml.md
@@ -86,7 +86,7 @@ The manifest file **MAY** contain zero or more subsections as defined below.
 
 #### Dependency Lists
 
-Dependencies lists contain all the resources that are required for this package, in various scenarios. Dependencies consist of a `key/value` pair where the `key` is a string that provides a unique name for the package in this project. This `key` can be either the qualified package name from the package's manifest, or an <a href="#alias">alias</a> for the package.
+Dependencies lists contain all the resources that are required for this package, in various scenarios. Dependencies consist of a `key/value` pair where the `key` is a string that provides a unique name for the package in this project. This `key` can be either the qualified package name from the package's manifest, a short-hand legacy reference (a bare package name e.g. `loadash` will expand out to `legacy@<current registry>/lodash`) or an <a href="#alias">alias</a> for the package.
 
 When the `key` value is the qualified package name (e.g. `toddself@static-pkg.dev/package2toml`), then the `value` **MAY** be either a semver range for the package or the <a href="#alias">alias</a> definition.
 
@@ -105,7 +105,7 @@ Shorthand: `"name@version"`
 
 The object **MAY** also contain:
 
-- `"patch"` - a `key`/`value` pair that allows specific overrides in the dependencies of the dependency. e.g. if this package depends on `lodash`, you could specifiy that you want it to use `underscore` instead: `"patch": {"legacy@npmjs.com/lodash": "toddself@static-pkg.dev/underscore"}`
+- `"patch"` - a `key`/`value` pair that allows specific overrides in the dependencies of the dependency. e.g. if this package depends on `lodash`, you could specifiy that you want it to use `underscore` instead: `"patch": {"legacy@registry.entropic.dev/lodash": "toddself@static-pkg.dev/underscore"}`
 
 This section could be used to also specify platform exclusion/inclusion for some dependencies. Some thought might be needed to figure out how to make this workable across the wide range of computer platforms that implementations might be written for.
 
@@ -130,7 +130,7 @@ To alias a package and specify a patch:
 "name" = "toddself@static-pkg.dev/package2toml"
 "version" = "^1.0.0"
 [dependencies.package2toml.patch]
-"legacy@npmjs.com/lodash" = "toddself@static-pkg.dev/underscore@1.0.5"
+"legacy@registry.entropic.dev/lodash" = "toddself@static-pkg.dev/underscore@1.0.5"
 ```
 
 #### Dependency Types
@@ -139,7 +139,7 @@ You **MAY** declare dependency types for:
 
 - `[dependencies]` - these are the packages that are required to run this package. They will be installed automatically when this package is required by another package, or when you specifically install the dependencies for this package manually.
 - `[devDependencies]` - these are packages that are required to develop this package. They will be installed only when you install the dependencies for this package manually. If a package depends on the current package, they will **NOT** be installed as part of that dependency graph.
-- `[peerDependencies]` - these packages are rqeuired to run this package, however, they will not be installed and should be additionally depended on or installed by either the parent package or along side this package for development.
+- `[peerDependencies]` - these packages are required to run this package, however, they will not be installed and should be additionally depended on or installed by either the parent package or along side this package for development.
 - `[optionalDependencies]` - these packages are not required to run this software, but may provide additional options or features in this package. They must be required or installed by the parent package or along side this package for development.
 
 ### Example [TK].toml
--- a/docs/RFCs/package-toml.md
+++ b/docs/RFCs/package-toml.md
@@ -1,0 +1,96 @@
+# Package.toml format
+
+A proposal for a more formal Package.toml definition
+
+- Version: DRAFT
+- Date: 06/06/2019
+- Authors
+  - Todd Kennedy <todd@selfassembled.org>
+
+## Abstract
+
+This document is meant to describe the package manifest format for Entropic,
+a federated solution to package management. Entropic needs to have its own
+manifest file due to conflicts arising from the namespacing entropic uses.
+
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
+"SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be
+interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Copyright
+
+This document is copyright by its individual contributors, and available for
+re-use under [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)
+
+## Introduction
+
+Entropic's cannot share a manifest format with the traditional `package.json` format
+associated with the ECMAScript ecosystem for a number of reasons:
+
+1. Entropic is intended to allow for federated package management for more than just the ECMAScript ecosystem
+1. The `package.json` name specifier does not allow for the namespacing that entropic provides
+1. There are recognized deficiencies with using JSON as human-readable metadata format
+
+## Filename
+
+The filename of the manifest **MUST** be `Package.toml` and it **MUST** live in
+the root directory of the package's filesystem. The root of the package is defined
+as the closest parent directory that contains the manifest file, starting with
+the current directory.
+
+On case-insensitive filesystems, it **MAY** be presented as `package.toml` but care
+should be taken to preserve case when possible.
+
+## Format
+
+The format of this file is **REQUIRED** to be in [TOML](https://github.com/toml-lang/toml)
+version 0.5.0 or later as defined by the project.
+
+### Top-Level Data
+
+The manifest **MUST** contain the following top-level keys:
+
+- `"name"` - the canonical name of the package, as a string, including it's namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
+- `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
+- `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `Package.toml` file. e.g `"./src/index.js"`
+
+The top-level of the manifest **MAY** also contain:
+
+- TK (differentiation between CJS and ESM module types, other languages, etc)
+
+### Subsections
+
+The manifest file **MAY** contain zero or more subsections as defined below.
+
+#### `[dependencies]`
+
+The dependencies section **MUST** list all the resources that are required for
+this package to run. The dependencies **MUST** be declared as `"key" = "value"`
+pairs where the `key` is the package's canonical name (as from the top-level section
+of it's `Package.toml` file) and the `value` is a valid SEMVER version or range
+expression that points to a valid package.
+
+Examples:
+
+To specify a specific version:
+
+```
+"toddself@static-pkg.dev/package2toml" = "1.0.4"
+```
+
+To specify all versions within a given major revision:
+
+```
+"toddself@static-pkg.dev/package2toml" = "^1.0.0"
+```
+
+To specify any version
+
+```
+"toddself@static-pkg.dev/package2toml" = "*"
+```
+
+## Contributors
+
+- Chris Dickensen [@chrisdickensen](https://github.com/chrisdickensen)
+- CJ Silverio [@ceejbot](https://github.com/ceejbot)
--- a/docs/RFCs/package-toml.md
+++ b/docs/RFCs/package-toml.md
@@ -56,10 +56,14 @@ The manifest **MUST** contain the following top-level keys:
 The manifest **MAY** contain:
 
 - `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `Package.toml` file. e.g `"./src/index.js"`
+- `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), and will signal to the node interpreter to run this code an ES6 Module.
+- `"license"` - a valid [SPDX license](https://spdx.org/licenses/) for your code
+- `"description"` - a short description of your package
+- `"homepage"` - a URL for your project's homepage
+- `"author"` - who wrote this package
 
-The top-level of the manifest **MAY** also contain:
-
-- TK (differentiation between CJS and ESM module types, other languages, etc)
+These fields **MAY** or **MAY NOT** be validated by CLI tools or the server-side
+software, depending on the implmentation.
 
 ### Subsections
 
--- a/docs/RFCs/package-toml.md
+++ b/docs/RFCs/package-toml.md
@@ -53,7 +53,7 @@ The manifest **MUST** contain the following top-level keys:
 - `"name"` - the canonical name of the package, as a string, including it's namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
 - `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
 
-The manifest **MAY** contain:
+The manifest **MAY** contain the following fields:
 
 - `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `Package.toml` file. e.g `"./src/index.js"`
 - `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), and will signal to the node interpreter to run this code an ES6 Module.
@@ -61,6 +61,7 @@ The manifest **MAY** contain:
 - `"description"` - a short description of your package
 - `"homepage"` - a URL for your project's homepage
 - `"author"` - who wrote this package
+- `"repository"` - where does this code live
 
 These fields **MAY** or **MAY NOT** be validated by CLI tools or the server-side
 software, depending on the implmentation.
--- a/docs/RFCs/package-toml.md
+++ b/docs/RFCs/package-toml.md
@@ -57,7 +57,7 @@ The manifest **MAY** contain the following fields:
 
 - `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `Package.toml` file. e.g `"./src/index.js"`
 - `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), and will signal to the node interpreter to run this code an ES6 Module.
-- `"license"` - a valid [SPDX license](https://spdx.org/licenses/) for your code
+- `"license"` - a valid [SPDX license](https://spdx.org/licenses/) for your code. If a license is not provided then the code is considered to be `UNLICENSED`.
 - `"description"` - a short description of your package
 - `"homepage"` - a URL for your project's homepage
 - `"author"` - who wrote this package
@@ -65,6 +65,8 @@ The manifest **MAY** contain the following fields:
 
 These fields **MAY** or **MAY NOT** be validated by CLI tools or the server-side
 software, depending on the implementation.
+
+Software implementing the entropic federation protocols and `Package.toml` handling **MAY** complain about any of these optional fields as missing (`license` being an obvious candidate).
 
 ### Subsections
 
--- a/docs/RFCs/package-toml.md
+++ b/docs/RFCs/package-toml.md
@@ -59,7 +59,7 @@ version 0.5.0 or later as defined by the project.
 
 The manifest **MUST** contain the following top-level keys, in order to be published:
 
-- `"name"` - the canonical name of the package, as a string, including it's namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
+- `"name"` - the canonical name of the package, as a string, including its namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
 - `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
 
 These keys are **NOT REQUIRED** be present when interacting with the CLI locally (e.g. installing depedencies).
@@ -71,7 +71,7 @@ The manifest **MAY** contain the following fields:
 - `"license"` - a valid [SPDX license](https://spdx.org/licenses/) for your code. If a license is not provided then the code is considered to be `UNLICENSED`.
 - `"description"` - a short description of your package
 - `"homepage"` - a URL for your project's homepage
-- `"author"` - who wrote this package
+- `"author"` - who wrote this package. this is any valid email identifier, or an array of valid identifiers.
 - `"repository"` - where does this code live
 - `"directory"` - for monorepo development, where in the monorepo does this package live (see: [npm RFC 10](https://github.com/npm/rfcs/blob/latest/implemented/0010-monorepo-subdirectory-declaration.md))
 
@@ -94,7 +94,7 @@ When the `key` value is the qualified package name (e.g. `toddself@static-pkg.de
 
 ##### Aliases
 
-Aliases can be used to provide short-hand references to longer or unweildy package names. The only requirement is that they are valid package names and be unique.
+Aliases can be used to provide short-hand references to longer or unwieldy package names. The only requirement is that they are valid package names and be unique.
 
 They should point to sub-objects which **MAY** contain the following two keys or **MAY** be the short-hand representation of this data. One of these options **MUST** be chosen:
 
--- a/docs/RFCs/package-toml.md
+++ b/docs/RFCs/package-toml.md
@@ -52,6 +52,9 @@ The manifest **MUST** contain the following top-level keys:
 
 - `"name"` - the canonical name of the package, as a string, including it's namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
 - `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
+
+The manifest **MAY** contain:
+
 - `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `Package.toml` file. e.g `"./src/index.js"`
 
 The top-level of the manifest **MAY** also contain:
@@ -62,10 +65,10 @@ The top-level of the manifest **MAY** also contain:
 
 The manifest file **MAY** contain zero or more subsections as defined below.
 
-#### `[dependencies]`
+#### Dependency Lists
 
-The dependencies section **MUST** list all the resources that are required for
-this package to run. The dependencies **MUST** be declared as `"key" = "value"`
+Dependencies lists contain all the resources that are required for
+this package, in various scenarios. The dependencies **MUST** be declared as `"key" = "value"`
 pairs where the `key` is the package's canonical name (as from the top-level section
 of it's `Package.toml` file) and the `value` is a valid SEMVER version or range
 expression that points to a valid package.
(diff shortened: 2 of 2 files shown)


docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,96 @@
...
+The format of this file is **REQUIRED** to be in [TOML](https://github.com/toml-lang/toml)
+version 0.5.0 or later as defined by the project.
+
+### Top-Level Data
+
+The manifest **MUST** contain the following top-level keys:
+
+- `"name"` - the canonical name of the package, as a string, including it's namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
+- `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
+- `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `Package.toml` file. e.g `"./src/index.js"`
```
chrisdickinson: Should `entry` be optional?

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,96 @@
...
+
+To specify any version
+
+```
+"toddself@static-pkg.dev/package2toml" = "*"
+```
+
+## Contributors
+
+- Chris Dickensen [@chrisdickensen](https://github.com/chrisdickensen)
```
chrisdickinson: s/kensen/kinson/g :smile:

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,96 @@
...
+
+The top-level of the manifest **MAY** also contain:
+
+- TK (differentiation between CJS and ESM module types, other languages, etc)
+
+### Subsections
+
+The manifest file **MAY** contain zero or more subsections as defined below.
+
+#### `[dependencies]`
```
chrisdickinson: `devDependencies`, `peerDependencies`, and `optionalDependendies` are also allowed.

docs/RFCs/package-toml.md
toddself: eek!

toddself:
Thanks for the feedback!

olingern:
Will **license** be supported inside of a `Package.toml`?

toddself:
@ceejbot and @olingern i updated the PR to include all those fields as optional except for `keywords`.  the keywords field has always felt extremely spammy to me and kind of unhelpful.  I'd rather figure out why to better index/weight search results (I've got a lot of ideas about search)

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,125 @@
...
+interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+
+## Copyright
+
+This document is copyright by its individual contributors, and available for
+re-use under [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)
+
+## Introduction
+
+Entropic's cannot share a manifest format with the traditional `package.json` format
```
zacanger: Entropic's, possessive?

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,125 @@
...
+associated with the ECMAScript ecosystem for a number of reasons:
+
+1. Entropic is intended to allow for federated package management for more than just the ECMAScript ecosystem
+1. The `package.json` name specifier does not allow for the namespacing that entropic provides
+1. There are recognized deficiencies with using JSON as human-readable metadata format
+
+## Filename
+
+The filename of the manifest **MUST** be `Package.toml` and it **MUST** live in
```
zacanger: There's some chatter about this in #147

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,125 @@
...
+The manifest **MUST** contain the following top-level keys:
+
+- `"name"` - the canonical name of the package, as a string, including it's namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
+- `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
+
+The manifest **MAY** contain:
+
+- `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `Package.toml` file. e.g `"./src/index.js"`
+- `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), and will signal to the node interpreter to run this code an ES6 Module.
+- `"license"` - a valid [SPDX license](https://spdx.org/licenses/) for your code
```
zacanger: Should `license` possibly be required?

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,125 @@
...
+- `"description"` - a short description of your package
+- `"homepage"` - a URL for your project's homepage
+- `"author"` - who wrote this package
```
zacanger: Should a `repository` or `source` optional field be included? That's something I find super valuable with other tools, and always miss it when it's not there.

docs/RFCs/package-toml.md
toddself: I dislike npm's requirement of it; a lot of code I write is for places I work which necessitates the unwieldy "UNLICENSED" to be typed and like man I typo that thing like a boss

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,125 @@
...
+
+These fields **MAY** or **MAY NOT** be validated by CLI tools or the server-side
+software, depending on the implmentation.
```
zacanger: `implmentation` -> `implementation`

docs/RFCs/package-toml.md
zacanger: Good point, I was thinking mostly of public FOSS packages, not the internal use case.

toddself: also prior art: `license` field in Python's setup.py, Rust's Cargo.toml, Ruby's gemspec all have license as optional

docs/RFCs/package-toml.md
toddself: good catch!

docs/RFCs/package-toml.md
ceejbot: I have to say, the repository field *is* super-useful, and it would be great if it were optionally present.

I will note that we don't have to get all the fields we want present just yet. If we get what we  know we need described now, we can treat this as a living document that we update as we go.

toddself: Added repo.  I thought about saying "other fields may appear here, go wild" but then i realized that we might want to not say that in case we need to reserve a field for a future usage that is incompatible with how people are  using it in the wild

docs/RFCs/package-toml.md
olingern: I think there's a discussion somewhere ( can't remember where ) about a license blocklist, i.e. _our project only allows MIT packages_

Personally, I think licensing is a super important part of installing a package. It's also a neat metric that you could run over a registry's database

chrisdickinson: We should make the `license` field optional, and on the server we can default it to `UNLICENSED` if not given. (& maybe yell about it in the CLI.)

toddself: I updated the spec to say that it'll default to `UNLICENSED` (which I believe is the tactic github takes on this), and that software implementing the entropic fedeverse may complain about any of these optional fields being missing (especially license)

nschonni:
Other reference docs from https://github.com/nodejs/package-maintenance might be useful (under the docs folder and the issue tracker). They've been discussing some of those "best practice" subjects when it comes to licensing, files, and versioning, that might be applicable here

toddself:
Also prior art:

* Ruby's [Gemspec](https://guides.rubygems.org/specification-reference/)
* Python's [Setup.py](https://docs.python.org/2/distutils/setupscript.html)
* Cargo's [Cargo.toml](https://doc.rust-lang.org/cargo/reference/manifest.html) (thanks @ceejbot)

dmfay:
Maven's [pom.xml](https://maven.apache.org/pom.html) format is a bit verbose and the build lifecycle aspect seems out of scope for Entropic, but it does a lot right with dependency resolution and management imo.

toddself:
> pom.xml
_takes long drag of cigarette_
> that's a filename i haven't heard in a long time

docs/RFCs/package-toml.md
zkat: I have serious concerns about this name, especially since [pnpm released support for .yaml and .json5](https://twitter.com/pnpmjs/status/1124733887056424963), and having them support `.toml` is not a stretch. The problems caused by case-normalizing filesystems like HFS+ (with default settings) and NTFS are indeed very serious and they, for example, can seriously mess with tooling like git.

All that said, I understand the Capitalized precedent of so many other modern package managers I can think of, and for that reason, I think having something _like_ this is a good idea.

I'd like to throw my hat in for `Entropic.toml`/`Entropic.lock` as the file pair that this spec uses, if only to distance it _further_ from being a JS-only dependency specifier, but also because a completely different name completely avoids any present or future conflicts in the name style.

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Subsections
+
+The manifest file **MAY** contain zero or more subsections as defined below.
+
+#### Dependency Lists
+
+Dependencies lists contain all the resources that are required for
+this package, in various scenarios. The dependencies **MUST** be declared as `"key" = "value"`
+pairs where the `key` is the package's canonical name (as from the top-level section
```
zkat: We have a unique opportunity here to move away from this restriction from the original `package.json` and add optional support for objects with extra metadata. This has long been a thorn on our side, and I'd love to see the injustice corrected.

This could _possibly_ translate to the disappearance of `[devDependencies]` and such, or make those categories optional in favor of, e.g. `"toddself@static-pkg.dev/package2toml" = { version: "1.0.4", dev: true, optional: true }`.

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,128 @@
...
+of it's `Package.toml` file) and the `value` is a valid SEMVER version or range
+expression that points to a valid package.
+
+Examples:
+
+To specify a specific version:
+
+```
+"toddself@static-pkg.dev/package2toml" = "1.0.4"
```
zkat: A thought comes to mind that these left-hand-side package names are _incredibly unwieldy_, and we'd be better served with leaving these long specifiers to the right-hand side. For example
```toml
[dependencies]
package2toml = "toddself@static-pkg.dev/package2toml@1.0.4"
```
This would instruct the package manager to make the name available as `package2toml` instead of `LONG@NAME/HERE`.

To riff off my earlier comment about objects, users might also choose to split these sections up:

```toml
[dependencies]
package2toml = { from = "toddself@static-pkg.dev/package2toml", version: "1.0.4" }
```

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,128 @@
...
+"toddself@static-pkg.dev/package2toml" = "*"
+```
+
+#### Dependency Types
+
+You **MAY** declare dependency types for:
+
+- `[dependencies]` - these are the packages that are required to run this package. They will be installed automatically when this package is required by another package, or when you specifically install the dependencies for this package manually.
+- `[devDependencies]` - these are packages that are required to develop this package. They will be installed only when you install the dependencies for this package manually. If a package depends on the current package, they will **NOT** be installed as part of that dependency graph.
+- `[peerDependencies]` - these packages are rqeuired to run this package, however, they will not be installed and should be additionally depended on or installed by either the parent package or along side this package for development.
```
zkat: insert generic whining that we need this for compatibility but I hate peerDeps and would love if we just redesigned them.

nschonni:
Maybe this is something to work out after the initial draft, but what about runtime/installtime platform dependencies. package.json had/has the old "engines" concept for specifying what the package would run on. I'm wondering if you've thought about being able to control what packages get pulled at install time, EX:
- I'm running on Windows, pull down this other package
- I'm running on Node 8<, so pull down an extra package, otherwise it's not needed

docs/RFCs/package-toml.md
zkat: another thought:

```toml
"toddself@static-pkg.dev/package2toml" = { version: "1.0.4", alias: "package2toml" }
```

but tbh I like the first example I gave much better.

zkat: Oh and from an implementation point of view, having a full-fledged specifier on the RHS makes it much easier to have this sort of thing coexist with existing specifiers (as in, https://github.com/npm/npm-package-arg)

docs/RFCs/package-toml.md
chee: i'm so into in this. leaving the future open for a package manager being able to handle more dependency environments than dev & prod. (for instance, i'd love to be able to specify which dependencies are the `test` dependencies)

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,128 @@
...
+
+The format of this file is **REQUIRED** to be in [TOML](https://github.com/toml-lang/toml)
+version 0.5.0 or later as defined by the project.
+
+### Top-Level Data
+
+The manifest **MUST** contain the following top-level keys:
+
+- `"name"` - the canonical name of the package, as a string, including it's namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
+- `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
```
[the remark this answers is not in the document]
zkat: See also: https://github.com/npm/read-package-json/issues/3

docs/RFCs/package-toml.md
toddself: yes! your insights are super helpful! thank you

docs/RFCs/package-toml.md
toddself: i added peerDeps because it's "supported" but god i hate peer deps so gosh darn much

toddself:
@nschonni i think that @zkat's ideas around how to [change the dependencies sections](https://github.com/entropic-dev/entropic/pull/189#discussion_r291767838) directly addresses those concerns. 

i've got band practice tonight but I plan on updating this Request for Comments with, well, all these comments!

docs/RFCs/package-toml.md
chrisdickinson: YES this. I'd also like to reserve the ability for applications to "open" a dependency to patch it, a la Cargo:

```toml
[dependencies."chris@registry.neversaw.us/foo".dependencies."lodash"]
patch = "chris@registry.neversaw.us/underscore"
```

docs/RFCs/package-toml.md
ceejbot: I am entirely in favor of `entropic.toml`, with a minor preference for lower case. Precedent & general convention win over my preference, though.

docs/RFCs/package-toml.md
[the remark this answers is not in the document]
ceejbot: Aha, yes. We were discussing the difference between things we'd require/allow in packages intended for publication vs packages that are end-users of dependencies: libraries vs apps. Requiring a name for libraries: yup. Requiring a name for apps: shrug. Another example: no git deps in things that are to be published, but have all the git deps you want in your unpublished application.

docs/RFCs/package-toml.md
ceejbot: https://discourse.entropic.dev/t/peer-deps-but-why-tho/24   <-- longer-form discussion, because I suspect we won't solve this problem in this PR.

docs/RFCs/package-toml.md
chrisdickinson: (We've discussing whether or not we want to support github-style deps – right now my M.O. is to only allow git deps when run at application level, for patching dependencies.)

docs/RFCs/package-toml.md
lynncyrin: Naming it `[P|p]ackage.toml` sounds like a good call if you're looking to keep CLI api compatibility with npm / yarn / etc. I don't believe that's one of your goals, though?

The first letter of a significant filename needing to be lowercase has made my life easier 100% of the times I've encountered it.

And as someone interested in entropic's model primarily for non-js packages (in the theoretical future where that's supported) - naming is `entropic.toml` (rather than package.toml, presumably for it's just connection) is also a bonus for me.

(aside: let me know if there's some pre-condition for commenting on this PRs, and if so I'll delete this)

docs/RFCs/package-toml.md
dmfay: The long names are a bit awkward, but doesn't assigning them to an arbitrary alias lead to potential weirdness with transitive dependencies? If I make lodash `_` and someone else makes it `lodash`, what happens when someone depends on both our libraries? Letting dependencies dictate what they're called seems like a sensible standard.

zkat: @chrisdickinson I like what Cargo does on this front: you're not allowed to publish using git deps but they're fine for local usage. Also yes, a patch feature would be nice.

@dmfay the package manager is able to differentiate those based on their full identifier + aliases. This is how the alias feature currently works in npm. I'm not worried about conflicts. The only downside here is that there might be less deduplication in the tree.

dmfay: I hadn't heard of an alias feature in npm! I didn't turn anything up in the docs, but I found [this](https://github.com/npm/npm/issues/2943) on the issue tracker which seems to be a conclusive 'no'. Could you elaborate?

I do think there's more at stake than duplication: it becomes less likely that copy+pasted code will "just work" if dependencies it references don't have to have a consistent name, which sounds more annoying the longer I think about it. And I don't know how it'd play with Symbols.

docs/RFCs/package-toml.md
eladb: Peer dependencies night also be the only possible type if Entropic will be used to manage packages for ecosystems which do not support multiple versions of the same dependency in their closures (most of them)

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,128 @@
...
+The filename of the manifest **MUST** be `Package.toml` and it **MUST** live in
+the root directory of the package's filesystem. The root of the package is defined
+as the closest parent directory that contains the manifest file, starting with
+the current directory.
+
+On case-insensitive filesystems, it **MAY** be presented as `package.toml` but care
+should be taken to preserve case when possible.
+
+## Format
+
```
eladb: Curious about the decision to use TOML instead of JSON. I’ve seen tremendous value in the ability to parse package.json files “natively” in node. Maybe worth describing the motivation for this decision as I am sure it will be a common question.

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,128 @@
...
+The manifest **MAY** contain the following fields:
+
+- `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `Package.toml` file. e.g `"./src/index.js"`
+- `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), and will signal to the node interpreter to run this code an ES6 Module.
+- `"license"` - a valid [SPDX license](https://spdx.org/licenses/) for your code. If a license is not provided then the code is considered to be `UNLICENSED`.
+- `"description"` - a short description of your package
+- `"homepage"` - a URL for your project's homepage
+- `"author"` - who wrote this package
+- `"repository"` - where does this code live
```
eladb: Location within the repo for mono-repos would be useful.

Generally mono-repos and large projects should get some attention. For example, being able to reuse package definitions across multiple packages. Happy to provide mode context when it’s relevant.

docs/RFCs/package-toml.md
zkat: Sorry, it's my bad. Aliases are... not really documented, and it's a fairly recent feature.

tl;dr you can do:

```json
"dependencies": {
  "underscore": "npm:lodash@1.2.3"
}
```

and it'll install `lodash` under `node_modules/underscore`.

docs/RFCs/package-toml.md
zkat: 👍 . See [the related RFC for npm](https://github.com/npm/rfcs/blob/latest/implemented/0010-monorepo-subdirectory-declaration.md)

docs/RFCs/package-toml.md
cmmartti: Comments, I expect. I've sometimes wanted to add a comment explaining why a dependency is pinned at a specific version, or add a description to a "scripts" command, but JSON doesn't allow that.

You could go with something like [JSON5](https://json5.org) but then you lose the ability to parse it "natively" in Node, and at that point it doesn't really matter what you use.

zacanger: There was some discussion about why TOML in https://github.com/entropic-dev/entropic/issues/106 and https://github.com/entropic-dev/entropic/issues/162, in particular https://github.com/entropic-dev/entropic/issues/106#issuecomment-497982913 and https://github.com/entropic-dev/entropic/issues/162#issuecomment-499264287

cmmartti: Those issues don't really explain why TOML was chosen, only why Entropic doesn't use package.json. Although I support the use of TOML, there's no reason the package file couldn't be entropic.json instead.

toddself: JSON doesn't support comments which has been a frequent complaint about it (especially when used for manifests). I'm also adding a "why TOML" section to the RFC

toddself:
> Maybe this is something to work out after the initial draft, but what about runtime/installtime platform dependencies. package.json had/has the old "engines" concept for specifying what the package would run on. I'm wondering if you've thought about being able to control what packages get pulled at install time, EX:
> 
>     * I'm running on Windows, pull down this other package
> 
>     * I'm running on Node 8<, so pull down an extra package, otherwise it's not needed

I think we can handle this with the expanded object scope

toddself:
Link to discourse discussion: https://discourse.entropic.dev/t/the-toml-file-lets-talk-about-it/23/17

toddself:
So far I've resolved all the conversations in this PR except:

* what to call the file itself
* peerDependencies

Please make new comments if I haven't captured your intent correctly! (There was a _lot_ of **excellent, positive** feedback!!!)

zacanger:
Maybe peerDeps could just be a future conversation and possible feature?

toddself:
A few more thoughts!

I think once we decide on the name for the file (which seems to be leaning towards`entropic.toml` and `lock.entropic` I think we can punt on the `peerDeps` issue until later, so ideally we can get this merged in and start building support for all these wonderful features

toddself:
Looks like @zacanger was 42 seconds faster with this thought!

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+## Format
+
+The format of this file is **REQUIRED** to be in [TOML](https://github.com/toml-lang/toml)
+version 0.5.0 or later as defined by the project.
+
+### Top-Level Data
+
+The manifest **MUST** contain the following top-level keys, in order to be published:
+
+- `"name"` - the canonical name of the package, as a string, including it's namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
```
ceejbot: it's ➜ its

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+- `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
+
+These keys are **NOT REQUIRED** be present when interacting with the CLI locally (e.g. installing depedencies).
+
+The manifest **MAY** contain the following fields:
+
+- `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `[TK].toml` file. e.g `"./src/index.js"`.
+- `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), and will signal to the node interpreter to run this code an ES6 Module.
+- `"license"` - a valid [SPDX license](https://spdx.org/licenses/) for your code. If a license is not provided then the code is considered to be `UNLICENSED`.
```
ceejbot: I really like this, says somebody who hates being yelled at by tooling about deficiencies  in package.jsons only ever used for installing deps...

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+- `"description"` - a short description of your package
+- `"homepage"` - a URL for your project's homepage
+- `"author"` - who wrote this package
```
ceejbot: This probably needs a note on valid/accepted formats. `"Basil Disco Smith <twelve@tardis.org>"` is an example of a valid author string. Also, you know what? Multiple authors are a thing we might want.

This comment should *not* block landing this PR. This is a living document we should feel free to change as needed until we release 1.0.

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+
+Dependencies lists contain all the resources that are required for this package, in various scenarios. Dependencies consist of a `key/value` pair where the `key` is a string that provides a unique name for the package in this project. This `key` can be either the qualified package name from the package's manifest, or an <a href="#alias">alias</a> for the package.
+
+When the `key` value is the qualified package name (e.g. `toddself@static-pkg.dev/package2toml`), then the `value` **MAY** be either a semver range for the package or the <a href="#alias">alias</a> definition.
+
+<a name="alias"></a>
+
+##### Aliases
+
+Aliases can be used to provide short-hand references to longer or unweildy package names. The only requirement is that they are valid package names and be unique.
```
ceejbot: unweildy ➜ unwieldy

Hello! I am your line editor for this evening. May I pour you some sparkling typo corrections?

docs/RFCs/package-toml.md
toddself: so i punted on this a little and just said "valid email identifiers or an array of them"

docs/RFCs/package-toml.md
toddself: this is literally the source of this.  the number of bullshit descriptions i've added  to packages...

zkat:
I think it's worth acknowledging that wherever this lands, it'll only be a starting point for more earnest development of a manifest spec that evolves as the client is developed. We can probably expect a lot of changes, so my inclination is to not get too hung up on little details, but probably just settle on the name and the basic dep specification stuff and otherwise this is a v good doc to start from <3

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+This document is copyright by its individual contributors, and available for
+re-use under [CC BY 4.0](https://creativecommons.org/licenses/by/4.0/)
+
+## Introduction
+
+Entropic cannot share a manifest format with the traditional `package.json` format
+associated with the ECMAScript ecosystem for a number of reasons:
+
+1. Entropic is intended to allow for federated package management for more than just the ECMAScript ecosystem
+1. The `package.json` name specifier does not allow for the namespacing that entropic provides
```
MylesBorins: can you expand on why this is the case?

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+
+The project has chosen to standardize on TOML as a format due to various short-comings in the JSON file specification, especially concerns around:
+
+1. Lack of comments
+1. Lack of multiline string support
+1. Designed for ease of generation by machine, not by hand
+
+## Filename
+
+The filename of the manifest **MUST** be `[TK].toml` and it **MUST** live in
```
MylesBorins: Can a package have multiple roots? That is currently the case for projects in the Node.js ecosystem where it is possible to have nested projects

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+## Format
+
+The format of this file is **REQUIRED** to be in [TOML](https://github.com/toml-lang/toml)
+version 0.5.0 or later as defined by the project.
+
+### Top-Level Data
+
+The manifest **MUST** contain the following top-level keys, in order to be published:
+
+- `"name"` - the canonical name of the package, as a string, including its namespace. e.g. `"toddself@static-pkg.dev/packge2toml"`
```
MylesBorins: is the idea that the canonical name of imported packages will also include the namespace?

e.g. `import 'toddself@static-pkg.dev/packge2toml'`

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+- `"version"` - a [SEMVER](https://semver.org/) designation conforming to version 2.0.0 of the semantic versioning specification. Due to how TOML parses numbers, this must be represented as a string. e.g. `"1.0.4"`
+
+These keys are **NOT REQUIRED** be present when interacting with the CLI locally (e.g. installing depedencies).
+
+The manifest **MAY** contain the following fields:
+
+- `"entry"` - the filename and package-relative path, as a string, to the main entry point of the package. The root for the relative-path is the directory which contains the `[TK].toml` file. e.g `"./src/index.js"`.
```
MylesBorins: is there a reason to stray from `main` which has canonically been used in the node.js ecosystem? This seems like it would create an extra source of confusion especially if it needs to be translated to main in a toml -> json conversion

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+- `"type"` - if set to `"module"` this will output a [`package.json` with this included](https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field), and will signal to the node interpreter to run this code an ES6 Module.
```
MylesBorins: the type doesn't actually signal to the loader what type to load, it signifies how to interpret file extensions

You can have `pkg.type` unset, and it would default to `commonjs`. Based on this current behavior having no `type` field and a `main` with a `.mjs` extension would still export an esm module.

Similarly a `type` of `module` with a `main` with the `.cjs` extension will still export a cjs module

docs/RFCs/package-toml.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+#### Dependency Types
+
+You **MAY** declare dependency types for:
+
+- `[dependencies]` - these are the packages that are required to run this package. They will be installed automatically when this package is required by another package, or when you specifically install the dependencies for this package manually.
+- `[devDependencies]` - these are packages that are required to develop this package. They will be installed only when you install the dependencies for this package manually. If a package depends on the current package, they will **NOT** be installed as part of that dependency graph.
+- `[peerDependencies]` - these packages are rqeuired to run this package, however, they will not be installed and should be additionally depended on or installed by either the parent package or along side this package for development.
+- `[optionalDependencies]` - these packages are not required to run this software, but may provide additional options or features in this package. They must be required or installed by the parent package or along side this package for development.
```
MylesBorins: I've heard requests for better ways of handling optional dependencies. E.g. how do I not install / compile brotli on newer versions of node.

docs/RFCs/package-toml.md
toddself: We can't support some of the options that we'd like to while keeping with a legacy format. in addition certain valid specifiers for entropic are [not valid for legacy registry clients](https://github.com/entropic-dev/entropic/issues/106#issuecomment-497982913)

docs/RFCs/package-toml.md
toddself: Yes, there will be support for monorepo (ncluding the location of this package within the monorepo).  

This should work like it does with the existing package manager.

docs/RFCs/package-toml.md
toddself: IIRC yes, but @chrisdickinson or @ceejbot might have a better answer for this.

docs/RFCs/package-toml.md
toddself: good point!

docs/RFCs/package-toml.md
toddself: clarified!

docs/RFCs/package-toml.md
toddself: Yeah, i think we have some definite work to do here in the deps section, especially with regard to the notably absent lifecycle scripts...

docs/RFCs/package-toml.md
chrisdickinson: Yes. Additionally, I was thinking about making packages available by the "smallest unique" name at each level too – so if there's no other `package2toml` at a given level, you could require it via `require('package2toml')`. However, that comes with a cost – so we deferred that behavior until later.

docs/RFCs/001-manifest-toml.md
```diff
@@ -0,0 +1,166 @@
...
+
+Software implementing the entropic federation protocols and `[TK].toml` handling **MAY** complain about any of these optional fields as missing (`license` being an obvious candidate).
+
+### Subsections
+
+The manifest file **MAY** contain zero or more subsections as defined below.
+
+#### Dependency Lists
+
+Dependencies lists contain all the resources that are required for this package, in various scenarios. Dependencies consist of a `key/value` pair where the `key` is a string that provides a unique name for the package in this project. This `key` can be either the qualified package name from the package's manifest, or an <a href="#alias">alias</a> for the package.
```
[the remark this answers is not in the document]
chrisdickinson: (Ah! And I just remembered: there is a shorthand syntax at play for keys: if you provide just `"lodash"`, that'll expand to `"legacy@<current registry>/lodash`)

docs/RFCs/package-toml.md
zkat: requiring using the fully qualified name feels like a big UX fail, imo. I think there can/should be some rule enforcement that makes it the default -- see also Mastodon's `@foo` being context-aware, so you don't have to type out someone's fully-qualified name pretty much -ever-.

olingern: Maybe it's possible for us to only require the package name by symlink'ing the downloaded package into the`node_modules` folder. 

So, `ds` would install `toddself@static-pkg.dev/packge2toml` as `package2toml`. I think this becomes more complex and simply importing the fully qualified name seems less error prone.

zkat: @olingern aliases solve this use case nicely

chrisdickinson: @zkat:

> requiring using the fully qualified name feels like a big UX fail, imo. I think there can/should be some rule enforcement that makes it the default -- see also Mastodon's `@foo` being context-aware, so you don't have to type out someone's fully-qualified name pretty much -ever-.

Yep, agreed – I want this feature (and aliases!) At the time, I deferred implementing the context-aware requirement until later because when I went to implement it, it seemed like it would have performance and deduping consquences that I wasn't prepared to tackle at the time. (Also: interactions with how legacy packages are unpacked vs. entropic packages.)

My goal was to produce a working system from which we could iterate. It seemed at the time like we could defer aliasing and context-aware import support until later, while reserving the right to re-define our existing dependency item format as a shorthand for an expanded one that supports those features down the line. (I was starting to get the "I'm too far out on a limb without supports" feeling from speccing aliasing and context-awareness without working from a backing implementation, if that makes sense?)

zkat: makes total sense to me :)

chrisdickinson:
We'll treat this as a living document & continue to iterate and fill things in as we get to them, but this is a great place to start from: THANK YOU!

static-raccoon:
This pull request has been mentioned on **Entropic**. There might be relevant details there:

https://discourse.entropic.dev/t/the-toml-file-lets-talk-about-it/23/1

static-raccoon:
This pull request has been mentioned on **Entropic**. There might be relevant details there:

https://discourse.entropic.dev/t/lifecycle-scripts-and-native-binaries/73/1

PR #193, opened 2019-06-07
feat: add routes for namespace creation/deletion

### LINKED ISSUE #113: implement namespaces workflow

superhawk610:
## Description

At present, the only method of creating namespaces outside of directly editing the database is by creating new user accounts. This PR adds routing to the server to handle creation and deletion of namespaces apart from user accounts.

## Notes

Closes #113.

### DIFF
--- a/services/registry/decorators/can-write-namespace.js
+++ b/services/registry/decorators/can-write-namespace.js
@@ -1,0 +1,59 @@
+'use strict';
+
+const Namespace = require('../models/namespace');
+const NamespaceMember = require('../models/namespace-member');
+const { response } = require('boltzmann');
+
+module.exports = canWrite;
+
+function canWrite(next) {
+  // is there a current user?
+  // does the namespace exist?
+  // if so, is the active user a member?
+
+  return async (context, params) => {
+    const { host, namespace } = params;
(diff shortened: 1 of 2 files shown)


superhawk610:
@chrisdickinson should it be possible to rename a namespace?

chrisdickinson:
@superhawk610 I tend to think it shouldn't be, since namespaces form part of a packages name & this would implicitly delete a bunch of packages (i.e., if I had a package `chris@registry.entropic.dev/semicolons` and changed my name to `isntitvacant`, the new package would be `isntitvacant@registry.entropic.dev/semicolons`.)

The rules are, I think:

- Namespaces can only be deleted if they have no packages
- Namespaces can only be deleted if they have a single member
- Namespaces may not remove their last active member
- Namespaces cannot be renamed, but new namespaces can be created and packages can be maintained by the new namespace.

cc @ceejbot to double check me on that.

(We should be unblocked now that #135 is merged now, too! Sorry for the delays)

superhawk610:
I don't have the time to grok the changes to the repo since I last worked on this so I'll have to close this out and let someone else take this. Best of luck on the project!

PR #196, opened 2019-06-07
docs: point to Arch Linux package

dmfay:
hi! I'm just starting to check this out but I thought I'd install The Arch Way™, so here's a package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,8 @@ Entropic requires a new command-line client, called `ds` (or "entropy delta".) *
 curl -sSL https://www.entropic.dev/install.sh | bash
 ```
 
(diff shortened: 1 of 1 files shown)


olingern:
Nice. I'm on arch, too. I'll test this out

toddself:
You'll need a different node version -- I don't believe this will under node 6.

dmfay:
@toddself I'm not following -- Arch ships [node 11.15.0](https://www.archlinux.org/packages/community/x86_64/nodejs/) right now, so you'd have to go outside its package management ecosystem to be running 6.

toddself:
Ah sorry - the deps list has `nodejs6-bin` listed

image

But it's that annoying substitutes list.

PR #200, opened 2019-06-07
Add nodemon to registry dev command

olingern:
### Description
Currently after running `docker-compose up` and then making changes to the registry, files do not autoreload.

This PR:
- Adds nodemon for autoreloading the registry
- Updates the registry start command to `npm run dev` in `docker-compose`

### How to test
- Pull this branch down: `git fetch origin pull/200/head:pr-200` ( where `origin` is your `entropic` remote )
- Run `docker-compose up` in the root ( don't daemonize if you want to see the logs! )
- Make a change to a registry file and watch reload.


olingern:
Not using the `pretest` hook gives a slight speed up on startup time, so that's also nice

PR #204, opened 2019-06-07
feat(ops): Kuberenetes

### LINKED ISSUE #116: Kubernetes support

zacanger:
Closes #116 
Depends on #220 

Open to PRs/patches! I'm still very new to Kubernetes, so please shout at me if you see any issues.

This is tested with microk8s. If anyone would be interested in testing on more production-like infrastructure I'd definitely welcome that. Otherwise, it should be ready to merge.

As mentioned in the comments, it would be neat to have a Helm chart, but with the Helm 2 to Helm 3 churn right now, I think that's something for the future.

### DIFF
--- a/misc/k8s/index.js
+++ b/misc/k8s/index.js
@@ -1,0 +1,26 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const { resolve } = require('path');
+const buildDir = resolve(__dirname, 'build');
+
+// clean up old files
+fs.readdirSync(buildDir)
+  .filter(f => f.endsWith('.json'))
+  .forEach(f => {
+    fs.unlink(`${buildDir}/${f}`, err => {
+      if (err) {
+        throw err;
+      }
+    });
+  });
+
+// write new json files for kubectl
+fs.readdirSync(resolve(__dirname, 'templates'))
+  .filter(f => f.endsWith('.js'))
+  .forEach(file => {
+    fs.writeFileSync(
+      resolve(buildDir, file.replace(/\.js$/, '.json')),
+      JSON.stringify(require(`./templates/${file}`), null, 2)
+    );
+  });
--- a/misc/k8s/templates/redis-deployment.js
+++ b/misc/k8s/templates/redis-deployment.js
@@ -1,0 +1,31 @@
+module.exports = {
+  apiVersion: 'extensions/v1beta1',
+  kind: 'Deployment',
+  metadata: {
+    creationTimestamp: null,
+    labels: { 'entropic-service': 'redis' },
+    name: 'redis'
+  },
+  spec: {
+    replicas: 1,
+    strategy: {},
+    template: {
+      metadata: {
+        creationTimestamp: null,
+        labels: { 'entropic-service': 'redis' }
+      },
+      spec: {
+        containers: [
+          {
+            image: 'redis:alpine',
+            name: 'redis',
+            ports: [{ containerPort: 6379 }],
+            resources: {}
+          }
+        ],
+        restartPolicy: 'Always'
+      }
+    }
+  },
+  status: {}
+};
--- a/misc/k8s/templates/storage-deployment.js
+++ b/misc/k8s/templates/storage-deployment.js
@@ -1,0 +1,64 @@
+module.exports = {
+  apiVersion: 'extensions/v1beta1',
+  kind: 'Deployment',
+  metadata: {
+    creationTimestamp: null,
+    labels: { 'entropic-service': 'storage' },
+    name: 'storage'
+  },
+  spec: {
+    replicas: 1,
+    strategy: { type: 'Recreate' },
+    template: {
+      metadata: {
+        creationTimestamp: null,
+        labels: { 'entropic-service': 'storage' }
+      },
+      spec: {
+        containers: [
+          {
+            args: ['npm', 'start'],
+            env: [
+              'CACHE_DIR',
+              'DEV_LATENCY_ERROR_MS',
+              'EXTERNAL_HOST',
+              'NODE_ENV',
+              'OAUTH_GITHUB_CLIENT',
+              'OAUTH_GITHUB_SECRET',
+              'OAUTH_PASSWORD',
+              'PGDATABASE',
+              'PGHOST',
+              'PGUSER',
+              'SESSION_EXPIRY_SECONDS',
+              'SESSION_SECRET',
+              'STORAGE_API_URL'
+            ].map(name => ({
+              name,
+              valueFrom: {
+                configMapKeyRef: {
+                  key: name,
+                  name: 'entropic-env'
+                }
+              }
+            })),
+            image: 'entropicdev/storage:latest',
+            imagePullPolicy: 'Always',
+            name: 'storage',
+            ports: [{ containerPort: 3000 }],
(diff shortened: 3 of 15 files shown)


krainboltgreene:
FWIW: https://twitter.com/kelseyhightower/status/963413508300812295?lang=en

zacanger:
Oh yeah, I definitely know. I think running postgres, redis, etc. in kubernetes should be an option here, but we should also allow it to work with externally hosted services (like RDS).

krainboltgreene:
One solution I've come up with for projects is to instead provide a
pgbouncer+pgpool interface. The application treats it like pg,so users
don't care, then they just have to hook up a postgres instance.

It still doesn't "scale" like kube expects to scale, but at least it
doesn't delete data when you scale 0.

zacanger:
I'll be honest, I'm a total postgres noob, but I'll take your word for it. Is that something we should consider adding?

krainboltgreene:
You want pgbouncer if you expect a lot of connections. You want pgpool if
you have a LOT more reads than writes.

zacanger:
@krainboltgreene I ended up removing Postgres from this entirely, since that seems like a decision admins should make (I know I'll end up just using RDS).

zacanger:
There've been a lot of changes today, so I need to give this another full test both locally and in my test cluster, but I think it's done enough to ask for a review.

misc/k8s/templates/storage-deployment.js
```diff
@@ -0,0 +1,64 @@
...
+                  key: name,
+                  name: 'entropic-env'
+                }
+              }
+            })),
+            image: 'entropicdev/storage:latest',
+            name: 'storage',
+            ports: [{ containerPort: 3000 }],
+            resources: {},
+            // volumeMounts: [{ mountPath: '/services', name: 'storage-claim' }],
```
zacanger: Last TODO that I know of.

zacanger:
That's pretty much how I feel about it. I am not a DBA and don't want to be one, and (maybe I'll end up being wrong about this, but) I don't think the target admin type user here is a DBA, or even an ops person. It's probably someone who cares about Node a lot, and may or may not have any database or operations experience.

krainboltgreene:
I know quite a bit about this surface of application development and I would love to do a pair review to transfer the knowledge.

krainboltgreene:
If I understand the misc/k8s directly I believe the current ecosystem leans towards using helm to do this job. Many many people will expect to be able to helm install entropic.

zacanger:
I thought about using Helm, but with Helm 3 coming soon I didn't know if it would be worth doing things twice. Once Entropic is more stable I think it would make sense to try to get charts into the official helm/charts repo, though.

krainboltgreene:
Agreed, and not having helm shouldn't stop this PR.

Julusian:
Why are you converting to yaml here? The Kubernetes api accepts json too, but does recommend yaml. As this project is really against yaml, why not just keep it as json?

https://kubernetes.io/docs/concepts/configuration/overview/
> Write your configuration files using YAML rather than JSON. Though these formats can be used interchangeably in almost all scenarios, YAML tends to be more user-friendly.

zacanger:
Good point. I was converting them to YAML because that's much more common, but it shouldn't really matter since it's already uncommon to write them in JS. I think I'll change that.

zacanger:
Some things took priority this week, but I should have this finished up this weekend.

.prettierrc
```diff
@@ -1,3 +1,5 @@
 {
-  "singleQuote": true
+  "singleQuote": true,
```
zacanger: Added this because of inconsistent Prettier behavior without it.

.gitignore
```diff
@@ -4,7 +4,7 @@ node_modules
 coverage
 legacy-cache
 allowed-list
-.DS_STORE
+.DS_Store
```
zacanger: Whoops (though I guess it might not matter, since Macs are funny about case)

misc/k8s/index.js
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const { resolve } = require('path');
+const buildDir = resolve(__dirname, 'build');
+
+// clean up old files
+fs.readdirSync(buildDir)
```
zacanger: Trying to avoid more deps

misc/k8s/templates/redis-deployment.js
```diff
@@ -0,0 +1,31 @@
+module.exports = {
```
zacanger: This is optional, could also use elasticache or similar

services/storage/package.json
```diff
@@ -70,7 +70,7 @@
     "lint-fix": "prettier --write '**/*.js'",
     "migrate": "PGUSER=postgres PGDATABASE=entropic_${NODE_ENV:-dev} db-migrate up --config db-migrate.json --env entropic",
     "posttest": "npm run lint",
-    "prestart": "NODE_ENV=dev npm run createdb >/dev/null; NODE_ENV=dev npm run migrate",
+    "prestart": "NODE_ENV=${NODE_ENV:-dev} npm run createdb >/dev/null; NODE_ENV=${NODE_ENV:-dev} npm run migrate",
```
zacanger: Sometimes we're in production!

misc/k8s/templates/queue-deployment.js
```diff
@@ -0,0 +1,40 @@
...
+      },
+      spec: {
+        containers: [
+          {
+            image: 'schickling/beanstalkd',
+            name: 'queue',
+            ports: [{ containerPort: 11300 }],
+            resources: {},
+            volumeMounts: [
+              { mountPath: '/var/lib/beanstalkd/', name: 'beanstalk-wal' }
```
alexindigo: Wondering how it works with scaling. Didn't have experience with mounted volumes in k8s. 
Btw, curious to hear why for Postgres was chosen different path.

zacanger: Right now this one actually doesn't do anything, there's no code there yet: https://github.com/entropic-dev/entropic/tree/master/services/workers. It might have to be changed in the future. I've tried pretty hard to avoid volumes in k8s in general, which is why I didn't even include a Postgres deployment. Running containers in databases with persistent volumes just seems like more pain than it's worth (at least to me)

alexindigo: I see. Thanks for the details. Yeah, persistent storage seems still to be terra incognita in k8s :)

zacanger:
@ceejbot @chrisdickinson @toddself (GH needs an @/moderators, @/maintainers, something like that) I did this in JS because writing out loads of yaml by hand is a PITA, but now I'm wondering if that might seem odd to someone who just wants to get this up and running. How do you feel about the JS generating JSON? It would only take a second to just change that to yaml, not sure which would be more approachable.

chrisdickinson:
Thanks for reviewing, @1stvamp! I've merged this in 6373162.

zacanger:
Thanks @1stvamp and @chrisdickinson!

PR #216, opened 2019-06-08
Add brew formula and move installation docs

zacanger:
I've never written one of these before, but it seems to work as expected. In the future when things are more stable it could be added to homebrew core.

Closes #196


moogacs:
I would like to do something like that, i already have  a PR about using wget as well and i think we should support different pkg managers

zacanger:
There's also #196 for Arch, maybe that should be split out into an `INSTALLING.md` at some point.

1stvamp:
Perhaps a more descriptive path name? e.g. `misc/homebrew/...`, then you could have `misc/pacman/...`, `misc/apt/...` etc.

zacanger:
@1stvamp I was looking at this as a temporary thing until `ds` is stable enough that we could just add the formula to homebrew core. I don't think hosting packages/pkgbuilds/etc here for every distribution would be the best long term solution

1stvamp:
From experience it generally makes it easier to release new packages for a
lot of systems (snaps, debs as immediate examples) if the package metadata
is part of the source, but I'm not going to die on this hill. :heart:

1stvamp:
@zacanger thank you for that 💖

zacanger:
@dmfay @moogacs not trying to steal your thunder, but if this gets merged we'd end up having conflicts so I brought your changes in here.

Closes #196 
Closes #185

misc/install.sh
```diff
@@ -1,6 +1,15 @@
...
+
+if hash curl 2>/dev/null; then
+  curl -SsL https://www.entropic.dev/ds-latest.tgz -o ds-latest.tgz
+elif hash wget 2>/dev/null; then
+  wget https://www.entropic.dev/ds-latest.tgz -o ds-latest.tgz
+else
+  echo 'Please install curl or wget!'
+  exit 1
+fi
+
```
ceejbot: This  reminds  me that we should have an automated way to push up  new versions of the client at some point. Also that I should hand over the demo site at some point.

PR #219, opened 2019-06-09
Reduce duplication in circle config

zacanger:
As more services and more tests are added, this could get unwieldy. This uses yaml "anchors" to reduce some of the duplication.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,18 @@
 version: 2
+
+setup: &setup
+  - checkout:
+      path: ~/repo
+  - restore_cache:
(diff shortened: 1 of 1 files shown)


zacanger:
I guess I'm less good at yaml than I thought (said everyone who's ever touched it).

1stvamp:
Problem is in circleci's clojure YAML parser is really finnickity about merging objects and lists using anchors, so you can't merge objects in, only lists. 

Anchors are handy for things like the top level config properties etc.

zacanger:
Ohh good to know

PR #221, opened 2019-06-09
Add lint and fail phase for root lint script.

samccone:
This should prevent any lint failing code to pass in presubmit prior to
the code being landed in source control.

---

Note: currently lint is not set to run during presubmit, however we can enable this! Doing this class of enable will be a `large-scale-change` by running prettier to reformat all code prior to enabling the lint to minimize code-churn for all contributors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,9 +25,10 @@
   "main": "index.js",
   "private": true,
   "scripts": {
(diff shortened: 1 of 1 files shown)


olingern:
#149 aimed to do this, but happy to close and let you take over!

samccone:
Ah apologies! I missed your change!
Closing mine in favor of yours.

Issue #211, opened 2019-06-08
switch from `url.parse()` to `new URL()`


ceejbot:
The older [node url api](https://nodejs.org/api/url.html#url_legacy_url_api) is deprecated. We should use the [newer WHATWG URL api](https://nodejs.org/api/url.html#url_constructor_new_url_input_base) instead.

Once https://github.com/entropic-dev/entropic/pull/135 lands we should do a sweep through the codebase and update. There's at least one url.parse usage added by PR #135.

moogacs:
only 1 exists currently , so will wait till #135  finished ?, I mean could be considered from now on to user `new URL` instead

ceejbot:
I just landed #135 so ... we might be down to only one? That's great! Easier to fix :)

moogacs:
@ceejbot , should be closed ?

zacanger:
For future reference, if you say `Fixes #211` or `Closes #211` in the PR description, it'll automatically close the associated issue.

moogacs:
what if I edited it now would have an effect ? or it has to be before the PR merged

zacanger:
It would have to be before the PR is merged

chrisdickinson:
We should be good now! Thanks @moogacs!

PR #223, opened 2019-06-09
Switch(#211): url.parse() to new URL()

moogacs:


services/common/boltzmann/router.js
```diff
@@ -15,9 +14,10 @@ function router(options) {
 
     return context => {
       const { request } = context;
-      const { pathname } = URL.parse(request.url);
+      const baseURL = 'http://' + request.headers.host + '/';
+      const { url } = new URL(request.url, baseURL);
```
ceejbot: @moogacs you  probably want `{ pathname }` here!

mirandaio:
@moogacs what about here? https://github.com/entropic-dev/entropic/blob/master/services/common/boltzmann/request-handler.js#L34

moogacs:
Hey @mirandaio  it's done now

PR #224, opened 2019-06-09
Fix minor spelling typo

samccone:
graceully => gracefully.

### DIFF
--- a/services/registry/server.js
+++ b/services/registry/server.js
@@ -35,7 +35,7 @@ const main = () => {
 
   // Docker gives containers 10 seconds to handle SIGTERM
   // before sending SIGKILL. Close all current connections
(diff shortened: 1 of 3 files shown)


services/registry/server.js
```diff
@@ -35,7 +35,7 @@ const main = () => {
 
   // Docker gives containers 10 seconds to handle SIGTERM
   // before sending SIGKILL. Close all current connections
-  // graceully and exit with 0.
+  // gracefully and exit with 0.
```
zacanger: i am so good at words

The same typo exists in storage/server.js and web/server.js, could you correct it there, too?

samccone: Done.

Issue #217, opened 2019-06-09
get the cli caught up with pr #135


ceejbot:
The namespace & maintainers endpoints changed in that PR, but the cli wasn't updated to use the new ones. 

See Chris's comment on the PR: https://github.com/entropic-dev/entropic/pull/135#issuecomment-500102315

olingern:
Cool if I take this one on? Should be able to have a working PR in the next day

PR #234, opened 2019-06-10
[CLI] - Catch cli up Pt. 1

olingern:
## Description
This is the first part of the large CLI refactor post #135 merge. The plan is to basically abstract fetch calls to a common API wrapper and mock those responses for easy testing. The wrapper will also be injected at runtime alongside the other options. This also adds `sinon` for spying / mocking / etc. 

`whoami` was just the easiest command that had the least amount of code to change to get this process started

Subsequent PRs will / should ( _potential good newcomer task?_ ) move the fetch from the command and into the API wrapper. Future PR's need to take heed to [this comment](https://github.com/entropic-dev/entropic/pull/135#issuecomment-500102315), specifically the URL changes

Related #217

## How to test
- Pull this branch down: `git fetch origin pull/234/head:pr-234` ( _where origin is your entropic remote_ )
- `cd` to `cli`
- Make sure you're logged in `node ./lib/main.js login`
- Run `node ./lib/main.js whoami`

You can link the `ds` to your current branch by running `npm link` in the `cli` directory, but I prefer to keep mine separate.

### DIFF
--- a/cli/lib/api.js
+++ b/cli/lib/api.js
@@ -1,0 +1,25 @@
+const fetch = require('node-fetch');
+
+class Api {
+  constructor(baseUrl, token) {
+    this.baseUrl = baseUrl;
+  }
+
+  headers(token) {
+    return {
+      authorization: `Bearer ${token}`
+    };
+  }
+
+  async whoami(token) {
+    return await fetch(`${this.baseUrl}/v1/auth/whoami`, {
+      headers: this.headers(token)
+    });
+  }
+
+  
+}
+
+module.exports = {
+  Api
+};
--- a/cli/lib/api.js
+++ b/cli/lib/api.js
@@ -1,7 +1,7 @@
 const fetch = require('node-fetch');
 
 class Api {
-  constructor(baseUrl, token) {
+  constructor(baseUrl) {
     this.baseUrl = baseUrl;
   }
 
--- a/cli/lib/api.js
+++ b/cli/lib/api.js
@@ -16,8 +16,6 @@ class Api {
       headers: this.headers(token)
     });
   }
-
-  
 }
 
 module.exports = {
--- a/cli/lib/api.js
+++ b/cli/lib/api.js
@@ -11,7 +11,7 @@ class Api {
     };
   }
 
-  whoami(token) {
+  whoAmI(token) {
     return fetch(`${this.baseUrl}/v1/auth/whoami`, {
       headers: this.headers(token)
     });
--- a/cli/lib/api.js
+++ b/cli/lib/api.js
@@ -11,13 +11,11 @@ class Api {
     };
   }
 
-  async whoami(token) {
-    return await fetch(`${this.baseUrl}/v1/auth/whoami`, {
+  whoami(token) {
+    return fetch(`${this.baseUrl}/v1/auth/whoami`, {
       headers: this.headers(token)
     });
   }
 }
 
-module.exports = {
-  Api
-};
+module.exports = Api;
--- a/cli/lib/commands/whoami.js
+++ b/cli/lib/commands/whoami.js
@@ -8,7 +8,8 @@ const { whoAmI } = require('../utils');
 const whoamiOpts = figgy({
   registry: { default: 'https://registry.entropic.dev' },
   token: true,
-  log: { default: require('npmlog') }
+  log: { default: require('npmlog') },
+  api: true
 });
 
 async function whoami(opts) {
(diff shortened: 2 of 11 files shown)


cli/lib/api.js
```diff
@@ -0,0 +1,21 @@
...
+  }
+
+  headers(token) {
+    return {
+      authorization: `Bearer ${token}`
+    };
+  }
+
+  whoami(token) {
+    return fetch(`${this.baseUrl}/v1/auth/whoami`, {
```
olingern: Maaaaybe we should move the version into a class variable.

cli/lib/commands/whoami.js
```diff
@@ -3,38 +3,34 @@
...
   try {
+    response = await opts.api.whoami(opts.token);
     body = await response.json();
   } catch (err) {
     opts.log.error(`Caught error requesting "${opts.registry}/v1/auth/whoami"`);
     return 1;
   }
 
-  if (response.status > 399) {
+  if (response.status !== 200 && response.status !== 204) {
```
olingern: Is it necessary to check for redirect code as the other conditional was?

cli/lib/main.js
```diff
@@ -26,17 +27,20 @@ async function main(argv) {
       }
     }
 
-    const registry =
-      args.registry ||
-      config.registry ||
-      env.registry ||
-      'https://registry.entropic.dev';
+    const registry = args.registry || config.registry || env.registry || 'https://registry.entropic.dev';
```
olingern: `lint-fix` brought this one in :thinking:

olingern:
Mmmk, `whoami` is working again. And then there was a logger, and it logged in color.

cli/lib/api.js
```diff
@@ -0,0 +1,21 @@
+const fetch = require('node-fetch');
+
+class Api {
+  constructor(baseUrl) {
+    this.baseUrl = baseUrl;
+  }
+
+  headers(token) {
```
enykeev: maybe then it makes sense to also load `.entropicrc` in API constructor and for module to export API instance you can copy\update rather than a class?

cli/lib/api.js
```diff
@@ -0,0 +1,21 @@
+const fetch = require('node-fetch');
+
+class Api {
```
chrisdickinson: Feel free to discard this suggestion if it is not useful, but have you thought about copy-pasting the internal storage client? The external and internal urls are _pretty similar_ and it might be a way to get a head start.

olingern: I could for sure. I was worried that it might be overwhelming to review as well as require a lot of rebasing since there's a lot of code moving around right now. What do you think?

cli/lib/commands/whoami.js
```diff
@@ -8,19 +8,19 @@ const { whoAmI } = require('../utils');
...
 
 async function whoami(opts) {
   opts = whoamiOpts(opts);
 
   try {
-    const username = await whoAmI();
-
-    console.log(username);
+    const username = await whoAmI(opts);
+    opts.log.log(username);
```
ceejbot: This ends up being very tidy, though I twitch at the log.log  repetition (I understand why it's there tho).

cli/lib/logger.js
```diff
@@ -0,0 +1,16 @@
...
+  }
+  static error(msg, stacktrace = undefined) {
+    console.log(chalk.default.red(msg));
+
+    // Don't print the stacktrace in red. That
+    // would be overwhelming
+    if (stacktrace) {
+      console.log(stacktrace);
+    }
+  }
```
ceejbot: This is a very nice hook spot for somebody to get fancy about output formatting.

cli/test/utils/FakeApi.js
```diff
@@ -0,0 +1,16 @@
+module.exports = class FakeApi {
```
ceejbot: A mock api client! I predict we'll pull this out into `common/` after it  grows some more features.

cli/lib/api.js
chrisdickinson: Go for what's easiest for now. The more I think about it, it's probably unlikely that it'd be helpful to duplicate the code, especially if there's a tricky rebase involved. Where there are API methods that are common to the internal and external clients we can make sure to keep those method names aligned as we add them, too.

cli/lib/commands/whoami.js
olingern: Agreed! I think it would be more idiomatic for it to read `opts.logger.log`, but that will require a larger PR because some parts of the code are currently using `opts.log.error`

PR #242, opened 2019-06-10
chore: cleanup pass

ceejbot:
I am methodically fixing things until `docker-compose up` produces a working system.


services/registry/.env-example
```diff
@@ -1,7 +1,7 @@
 NODE_ENV=dev
 DEV_LATENCY_ERROR_MS=10000
 REDIS_URL=redis://redis:6379
-PORT=3000
+PORT=3002
```
chrisdickinson: Noted via another channel: the expected port layout is 3000 for registry, 3001 for web, 3002 for storage.

docker-compose.yml
```diff
@@ -32,7 +32,7 @@ services:
     volumes:
       - ./services/:/services
     working_dir: /services/registry
-    command: sh -c "npm run prestart && npm run dev"
+    command: sh -c "npm run dev"
```
olingern: So, I changed this to `npm run dev` because I assumed that `npm start` would be reserved for production and `nodemon` shouldn't run in production. 

The downside was that `prestart` no longer ran because `npm start` no longer used.[

[Commit](https://github.com/entropic-dev/entropic/commit/7492f789fb48a0a51ad3006782cc57b866492661#diff-4e5e90c6228fd48698d074241c2ba760R35)

ceejbot: The service split-out means there is no longer a prestart for the registry, which doesn't have a db anymore! It's all in `storage` now. `npm run dev`  is the right choice.

PR #247, opened 2019-06-11
Chris/fix docker compose and url

chrisdickinson:
- Fixes `new URL`'s pickiness about needing a FQDN param if given a partial path.
- `new URL` spells `context.url.query.page` as `context.url.searchParams.get('page')`
- Add default docker environment values so that they all know about each other in docker-compose mode.
- Add an admonition that "we don't mean for you to deploy a production entropic using docker-compose."

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,5 @@
+# docker-compose is provided for development purposes only and is not
+# intended for use as a production entropic setup.
 version: '3.1'
 services:
   db:
@@ -39,6 +41,12 @@ services:
       - "3000:3000"
     env_file:
       - ./services/registry/.env
+    environment:
+      STORAGE_API_URL: http://storage:3002
+      PORT: 3000
+      WEB_HOST: http://localhost:3001
+      REDIS_URL: redis://redis:6379
+      EXTERNAL_HOST: http://localhost:3000
 
   web:
     image: mhart/alpine-node:12
@@ -52,6 +60,12 @@ services:
       - "3001:3001"
     env_file:
       - ./services/web/.env
+    environment:
+      STORAGE_API_URL: http://storage:3002
+      PORT: 3001
+      WEB_HOST: http://localhost:3001
+      REDIS_URL: redis://redis:6379
+      EXTERNAL_HOST: http://localhost:3001
 
   storage:
     image: mhart/alpine-node:12
@@ -66,6 +80,13 @@ services:
     #  - "3002:3002"
     env_file:
       - ./services/storage/.env
+    environment:
+      POSTGRES_URL: postgres://postgres@db:5432
+      PGHOST: db
+      REDIS_URL: redis://redis:6379
+      PORT: 3002
+      PGUSER: postgres
+      PGDATABASE: entorpic_dev
 
 volumes:
   postgres_data:
--- a/package.json
+++ b/package.json
@@ -52,8 +52,8 @@
     "lint-fix": "prettier --write '**/*.js'",
     "lint-md": "markdownlint \"**/*.md\" -i \"**/node_modules/**\"",
     "lint-registry": "cd services/registry; npm run lint",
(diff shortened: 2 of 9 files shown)


docker-compose.yml
```diff
@@ -1,3 +1,5 @@
+# docker-compose is provided for development purposes only and is not
+# intended for use as a production entropic setup.
```
ceejbot: Contact  your doctor if symptoms persist for more than three days.

docker-compose.yml
```diff
@@ -1,3 +1,5 @@
+# docker-compose is provided for development purposes only and is not
```
zacanger: This is an extremely good warning

services/common/boltzmann/request-handler.js
```diff
@@ -29,7 +29,7 @@ class Context {
     if (this._parsedUrl) {
       return this._parsedUrl;
     }
-    this._parsedUrl = new URL(this.request.url);
+    this._parsedUrl = new URL(this.request.url, 'http://entropic.dev');
```
ceejbot: Hrm, should be EXTERNAL_HOST, yes?

services/registry/handlers/namespaces.js
```diff
@@ -150,8 +150,8 @@ async function memberships(context, { user }) {
 async function maintainerships(context, { namespace, host }) {
   const [err, result] = await context.storageApi
     .listNamespaceMaintainerships({
-      page: Number(context.url.query.page) || 0,
-      status: context.url.query.status,
+      page: Number(context.url.searchParams.get('page')) || 0,
+      status: context.url.searchParams.get('status'),
```
ceejbot: I am sure there's a Good Reason™ why url.parse had to go away, leaving us with this less-nice API.

services/common/boltzmann/request-handler.js
chrisdickinson: I opted to go with an (inert) URL here because the host value is never used except to keep `new URL()` from throwing an exception; whereas this code is used from a lot of different services which may not _all_ need (or specify) `EXTERNAL_HOST`.

services/registry/handlers/namespaces.js
chrisdickinson: I will (grudgingly) cede that if you _wanted_ to port this entire thing to a service worker based system of some cloudflare-y variety, you'd have to make this change anyway.

services/common/boltzmann/request-handler.js
ceejbot: Oh good LORD.  Fine then.

docker-compose.yml
```diff
@@ -66,6 +80,13 @@ services:
     #  - "3002:3002"
     env_file:
       - ./services/storage/.env
+    environment:
+      POSTGRES_URL: postgres://postgres@db:5432
+      PGHOST: db
+      REDIS_URL: redis://redis:6379
+      PORT: 3002
+      PGUSER: postgres
+      PGDATABASE: entorpic_dev
```
Daniel15: Typo? ent**or**pic instead of ent**ro**pic 😛

package.json
```diff
@@ -52,8 +52,8 @@
     "lint-fix": "prettier --write '**/*.js'",
     "lint-md": "markdownlint \"**/*.md\" -i \"**/node_modules/**\"",
     "lint-registry": "cd services/registry; npm run lint",
-    "postinstall": "for d in cli services/{registry,workers,web,common/boltzmann}; do cd $d; npm i; cd -; done",
+    "postinstall": "for d in cli services/{registry,workers,web,storage,common/boltzmann}; do cd $d; npm i; cd -; done",
```
Daniel15: Hmm... I don't think this will work on Windows.

zacanger: I think it would with git-bash or WSL. There's a lot in this project that I don't think would work with cmd.exe or powershell, I don't know of a good way around that besides maybe adding links to docs on WSL in HACKING.md.

PR #259, opened 2019-06-11
fix(post-install): make top level post-install script compatible with POSIX shells like dash

1stvamp:
# Change Type

* [ ] Feature
* [ ] Chore
* [x] Bug Fix

# Change Level

* [ ] major
* [ ] minor
* [x] patch

POSIX shells like `dash` don't support curly brace expansion (e.g. `"foo/{bar,baz}"` expanding to `"foo/bar foo/baz"`), and most linuxes symlink `/bin/sh` to `dash` or other limited shells rather `bash` or `zsh` for security reasons.

If you run `npm ci` from the top level directory this will cause the `post-install` script to infinitely loop with this error scattered in the stderr output: `sh: 1: cd: can't cd to services/{registry,workers,web,storage,common/boltzmann}`

This fix uses `printf` and subshells to mostly keep the conciseness and readability of the original while remaining compatible with most shells, and also uses a subshell to negate the requirement to `cd -` after each loop iteration.

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,8 +52,8 @@
     "lint-fix": "prettier --write '**/*.js'",
     "lint-md": "markdownlint \"**/*.md\" -i \"**/node_modules/**\"",
     "lint-registry": "cd services/registry; npm run lint",
-    "postinstall": "for d in cli services/{registry,workers,web,storage,common/boltzmann}; do cd $d; npm i; cd -; done",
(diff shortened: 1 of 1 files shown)


zacanger:
I never would've thought to do that, nice.

toddself:
Thank you @1stvamp!

PR #261, opened 2019-06-12
cli/lib/main now rejects command names with file metachars

### LINKED ISSUE #251: possible parent path traversal in command argument
# Is this a feature request or a bug?
bug

Parent [path traversal](https://en.wikipedia.org/wiki/Directory_traversal_attack) is often considered a security problem.
This is likely low impact if it is a security problem; the only vectors I can see is socially engineered copy/paste or postinstall hook abuse.

If an attacker can get a JS file onto the machine, they might be able to use shell injection or malicious hooks to cause entropic to load that JS file, but an attacker can usually do anything via shell access that they could by running JS in the entropic process unless entropic evolved to accept multiple commands as a long-lived, privileged process.

# Actual behavior:
https://github.com/entropic-dev/entropic/blob/ca58c5e9e4a94873577e7ce2ee7745d4d540f867/cli/lib/main.js#L21

`entropic ../../../node_modules/prettier` fails with an error like "TypeError: ... is not a function"

This works because `../../..` backs out of `cli/lib/commands` to the root directory for entropic.

# Expected behavior:
A command passed as the zero-th argument at the command line should not cause line 21 to load a package outside the commands subdirectory.  Line 21 should not load a dev dependency or a globally installed non-dependency.

Specifically, `entropic ../../../node_modules/prettier` should dump help info instead of potentially calling prettier's export as an async function.

# Steps to replicate:
```sh
$ echo 'console.log("Hello, World")' > ./poc.js
$ entropic ../../../poc
Hello, World
```

# Possible fix

One way to address is to define a `function isCommandArgument` that uses `require.resolve` to convert `argv[0]` to a file path, then use `path.relative` to find a path relative to __DIRNAME and see if the first component of that relative path is `commands`.

Alternatively, if there is no desire to support commands like `subdir/basename` in the future then main could dump help text when `/[\/\\]/.test(argv[0])`.

I can put together a PR if desired.

mikesamuel:
Fixes #251

This adds a check to function main so that when the first argument,
which is interpreted as a command name, has a file separator, it is
rejected, logged, and the command `help` is used instead.

To ease testing, this splits the analysis code out of `function main`
into `function unpack` which find the `cmd` to invoke but leaves the
invoking and error recovery to `main`.

This adds tests that a dodgy command is rejected, but that others like
`whoami` continue to function.

```sh
$ (pushd cli; npm t; popd)
$ npm run lint
```

# Change Type

* [ ] Feature
* [ ] Chore
* [X] Bug Fix

# Change Level

* [ ] major
* [X] minor
* [ ] patch

# Checklist

### DIFF
--- a/cli/lib/main.js
+++ b/cli/lib/main.js
@@ -3,6 +3,7 @@
 'use strict';
 
 module.exports = main;
+module.exports.unpack = unpack;
 
 const minimist = require('minimist');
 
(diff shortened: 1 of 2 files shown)


chrisdickinson:
Merged in f14b325a371ad6c7fb5d85a132a015f49d3ed583, thanks!

Issue #262, opened 2019-06-12
Distribution


Daniel15:
How are you planning on distributing Entropic to users? Will people have to manually download it, or will you just distribute it via npm (and expect people to have npm installed), or something else entirely?

For what it's worth, for Yarn we provide a few options, like a Windows installer, Debian/RPM packages, an installation script, and also publish it to npm.

Sorry if this is a duplicate issue - I did skim the other issues but couldn't see anything related.

zacanger:
`ds` can currently be installed from the AUR, with Brew, and with curl or wget (docs [here](https://github.com/entropic-dev/entropic/blob/master/docs/installing/README.md)), and there's an issue for making it installable through Entropic itself [here](https://github.com/entropic-dev/entropic/issues/67).

Daniel15:
Cool, thanks, not sure how I missed that document. Would be good to have a way that works well for Windows users too, given [it's the most popular development OS](https://insights.stackoverflow.com/survey/2019#technology-_-developers-primary-operating-systems).

zacanger:
The curl and wget instructions should work for Windows as long as msysgit/WSL/something like that is set up, but in the future an MSI would probably be a good idea.

fetis:
> The curl and wget instructions should work for Windows as long as msysgit/WSL/something like that is set up, but in the future an MSI would probably be a good idea.

I can confirm it. I got `ds` installed on Windows with help of git once I downloaded the bash script and opened it.

fetis:
I think I'll update the installation guide with Windows section, since the fact that it actually works was a big surprise for me.

Daniel15:
So I just realised that PowerShell aliases `wget` to its built-in equivalent (`Invoke-WebRequest`), so `wget` works in PowerShell even if you don't actually have wget installed. Might be worth mentioning that? You'd actually need Bash to execute the script though, so perhaps that's not actually relevant.

PR #264, opened 2019-06-12
docs: Windows installation

fetis:
# Change Type

* [ ] Feature
* [x] Chore
* [ ] Bug Fix

# Change Level
n/a

Added Windows installation instructions as mentioned in https://github.com/entropic-dev/entropic/issues/262

### DIFF
--- a/docs/installing/README.md
+++ b/docs/installing/README.md
@@ -23,4 +23,4 @@ If you're using Arch Linux, install the [`nodejs-entropic`](https://aur.archlinu
 ### Windows
 
 If you have Git installed you can download [installation script](https://www.entropic.dev/install.sh) and execute it with double click. 
-There's no standalone MSI installer though at the moment.
+There's no standalone MSI installer at the moment.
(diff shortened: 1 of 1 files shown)


toddself:
I think we'd be better served saying you need some sort of "*nix on windows" environment set up.  The `.sh` works if you have git installed with msys (and would work with WSL as well).

Ideally we'd have a cross-platform installer that can execute correctly.

fetis:
@toddself 
Actually, there's no magic behind `install.sh` it just downloads the package and installs it.
I have just checked and maybe we can drop download phase in favor of 
`NPM_CONFIG_ENGINE_STRICT=true npm i -g https://www.entropic.dev/ds-latest.tgz`

docs/installing/README.md
```diff
@@ -18,3 +18,9 @@ brew install https://raw.githubusercontent.com/entropic-dev/entropic/master/docs
 ### Arch Linux
 
 If you're using Arch Linux, install the [`nodejs-entropic`](https://aur.archlinux.org/packages/nodejs-entropic/) package from the AUR.
+
+
+### Windows
+
+If you have Git installed you can download [installation script](https://www.entropic.dev/install.sh) and execute it with double click. 
+There's no standalone MSI installer though at the moment.
```
varjmes:
```suggestion
There's no standalone MSI installer at the moment.
```

reads better.

fetis:
@varjmes @toddself here's the new version including manual option and extended win info

PR #265, opened 2019-06-12
chore: gitignore for IDEs

fetis:
# Change Type

* [ ] Feature
* [x] Chore
* [ ] Bug Fix

# Change Level
n/a

- Added strings for JetBrains IDEs and VS Code
- fixed indentation in `cli/package.json`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -7,3 +7,10 @@ allowed-list
 .DS_Store
 *~
 *.swp
+services/data/entropic-cache
+.opt-in
+.opt-out
+
+# IDEs
+.idea
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ node_modules
 coverage
 legacy-cache
 allowed-list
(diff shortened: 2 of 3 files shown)


enykeev:
How about we use something like that instead https://www.gitignore.io/api/osx,node,intellij,visualstudio?

zacanger:
Whew, that's a lot of stuff to ignore @enykeev. A lot of those files won't ever show up in this project (I doubt anyone is using ReSharper or MSSQL in a Node project), so I think it might be better to just add things to the ignores files as they come up.

@fetis could you add those same lines to the `.dockerignore`?

fetis:
> How about we use something like that instead https://www.gitignore.io/api/osx,node,intellij,visualstudio?

I saw this generator and in my opinion it does too much. I would like to keep in ignore only required stuff.

fetis:
> could you add those same lines to the .dockerignore?

@zacanger synced with .gitignore

Issue #226, opened 2019-06-10
Ink for CLI


olingern:
[Ink](https://github.com/vadimdemedes/ink) is a super neat project that allows command line development using React-like components. I think using this _could_ make the CLI more approachable for newcomers as well as providing a nice layer of visual abstraction.

zacanger:
TIL. That looks _so_ cool and I already want to rewrite half of my own projects to use it.

olingern:
It's super interesting. [Yoga layout](https://yogalayout.com/) ( _how Ink positions_ ) turns out to be really useful.

chrisdickinson:
Selfishly, I've been wanting to use this so I'm all for it.

I'd kind of like to see the CLI separate into an API client layer (a la `common/boltzmann/client/index.js`), display layer (in ink, potentially) and dispatch/command layer (what we have now), with our various utility methods getting absorbed into the client or the command layer.

emkay:
Ink looks super rad and I've also been looking for an opportunity to try it out and can help out porting stuff over.

boneskull:
ink is neat but it isn’t without overhead. aside from the tooling, it will incur a noticeable startup penalty. maybe this is the heaviness of React itself—haven’t looked closely at where the bottleneck is.

joehand:
If anyone is interested in alternatives I will plug https://github.com/neat-log/neat-log which we use in the Dat command line. It is still somewhat minimal but I would love any critical but kind eyes on it.  I have also been eyeing Ink but have yet to compare more in depth...

Folks have written a chat client with it and mostly works =) https://github.com/cabal-club/cabal#cabal-cli

MichaelDimmitt:
Another alternative: [oclif](https://oclif.io/)

ChrisBrownie55:
@boneskull If the bottleneck is in React itself, I wonder if it would be viable to use Preact in its place?

boneskull:
@ChrisBrownie55 I don't think so.  Ink uses [yoga](https://github.com/facebook/yoga); preact does not.

ChrisBrownie55:
@boneskull I was referring to using Preact in place of React, this would still be in conjunction with Ink. But I've tried to test using `preact/compat` by aliasing `react` to it but it would appear that Ink relies too heavily on the internal systems/variables of React making impossible to just switch to Preact.

wardpeet:
[...]

PR #269, opened 2019-06-13
Catch up cli pt2

olingern:
## Description

Follow up to #234 

This PR aims to: 
- do some refactoring for test ability
- Create a "core" that isn't concerned with presentation, i.e. logging
- Move API construction to the api lib
- Use the injected logger instead of `console.log`
- Fix any breakage from #134 

**The idea**
I think the CLI can follow a sort of mvc style of structure, `core` being the model and the `commands` directory being the controller. We don't have a view at the moment, and it's not necessary to; however, it would be nice to have the option to explore #226 -- this would make that transition very easy.

**Currently**
I'm testing and pushing packages against `registry.entropic.dev` because that version appears to be pre-#134, which is great. After the refactoring is done, I'll just switch over to my local docker instance and find the breakage and update `api.js` accordingly.

## How to test

_coming soon_


olingern:
So, this PR is coming along pretty well. Isolating publish was / is tricky because of its intermediary messages, but I solved passing this between the command and core with a "message broker" / `EventEmitter`.

What do you think of the current direction @chrisdickinson? My idea is that entropic's core could be wrapped up into a programmatic lib for people to use if they like ( _in the distant future_ ), so I thought pulling things apart was worthwhile. This will also make #226 very straightforward

cli/test/lib/core/ping.js
```diff
@@ -0,0 +1,59 @@
+const test = require('ava');
+const FakeApi = require('../../utils/FakeApi');
+
+// Mock private function that uses Date.now to
+// get consistent return value
+const rewire = require('rewire');
```
olingern: This is kind of hacky. I might be able to find a way to do this with `sinon`, but didn't see a clear way to do this in their docs.

cli/test/utils/FakeApi.js
```diff
@@ -1,15 +1,22 @@
 module.exports = class FakeApi {
-  constructor(response, status) {
-    this.desiredResponse = response;
-    this.desiredStatus = status;
+  constructor(mockValueObj) {
+    Object.keys(mockValueObj).forEach(
```
olingern: I opted to rewrite this so that tests can build the class endpoints they like. So, given a key like, `whoami` it will bind a function to be invoked to the `FakeApi` class. 

While this doesn't document the api as much, it's much more flexible and handles things like `publish` where multiple responses need to be built.

cli/test/utils/broker.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+const { MessageBroker, MESSAGE_EVENTS } = require('../../lib/MessageBroker');
+
+module.exports = {
+  createFakeBroker
+};
+
+function createFakeBroker() {
```
olingern: currently only used in publish. this redirects events to be pushed to an array instead of the runtime console logger.

cli/lib/commands/publish.js
```diff
@@ -1,200 +1,31 @@
...
-      }
-    }
-  );
 
-  const body = await request.json();
-  if (!request.ok) {
-    opts.log.error(body.message || body);
-    return 1;
+  try {
+    const tomlObject = await loadPackageToml(process.cwd());
```
olingern: Injecting the toml object into publish made using a toml fixutre much easier

cli/lib/api.js
```diff
@@ -1,21 +1,112 @@
...
+    this.token = token;
     this.baseUrl = baseUrl;
+    this.baseUrlV1 = `${baseUrl}/v1`;
   }
 
-  headers(token) {
+  headers() {
     return {
-      authorization: `Bearer ${token}`
+      authorization: `Bearer ${this.token}`
```
olingern: I decided to inherit this from `main.js` since a user _should_ be logged in at runtime for a command that needs to be auth'd

cli/lib/MessageBroker.js
```diff
@@ -0,0 +1,47 @@
...
+};
+
+/**
+ * Produces an instance of MessageBroker and binds
+ * a logger to message events.
+ *
+ * @param {*} logger Inject a logger. Console, file,
+ *            etc.
+ */
+const createMessageBroker = logger => {
```
olingern: Basic factory to produce broker instances to listen for `INFO` and `ERROR` events

PR #270, opened 2019-06-13
Add snap pkg builds

1stvamp:
# Change Type

* [x] Feature
* [ ] Chore
* [ ] Bug Fix

# Change Level

* [ ] major
* [ ] minor
* [x] patch

This adds a build file (`snapcraft.yaml`) for building a [snap package](https://snapcraft.io/), which can be pushed up to the snap store (I have published an initial _edge channel_ release [here](https://snapcraft.io/entropic-ds) and would be happy to hand over ownership to current maintainers) and installed on any Linux that supports `snapd`.

~Following up, and given credentials for talking to the snap store, `.travis.yml` could be updated to continuously push `master` builds to the aforementioned edge channel.~
Actually you can add a GitHub integration (webhook) easily from snapcraft.io to do the continuous builds on Canonical's build farm and published continuously to the store. 🎉

# Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ allowed-list
 services/data/entropic-cache
 .opt-in
 .opt-out
+*.snap
--- a/cli/snapcraft.yaml
+++ b/cli/snapcraft.yaml
@@ -1,0 +1,27 @@
+---
+name: ds
+summary: "ds: entropy delta; the entropic client"
+description: |
+  A client for the federated entropy package registry,
+base: core18
+grade: devel
+adopt-info: ds
+
+parts:
+  ds:
+    override-pull: |
+      snapcraftctl pull
+      DS_SNAP_VERSION=${DS_SNAP_VERSION:-master}
+      snapcraftctl set-version "${DS_SNAP_VERSION}"
+    source: .
+    plugin: nodejs
+    nodejs-version: 12.4.0
+    nodejs-package-manager: npm
+
+apps:
+  ds:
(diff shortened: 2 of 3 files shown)


1stvamp:
I welcome feedback on where _how to build a snap package` instructions should live (the current `cli/README.md` seems appropriate except it currently has info purely about using the `ds` util, not about developing/packaging it for consumption).

For the record the usage for building a snap is this:

```sh
cd cli/
snapcraft

# This gives you a ds_master_amd64.snap file
# You can install and test it, --dangerous means ignore the lack of signature for the package that comes when publishing it to the store
snap install --dangerous ds_master_amd64.snap

# To publish it to the snap store
snapcraft login # unless already logged in
snapcraft push ds_master_amd64.snap

# If it's good release it to the edge channel
snapcraft release entropic-ds $REVISION edge

# It should show up momentarily in the store
snap find entropic-ds
```

zacanger:
Woah this is awesome. This is all it takes to build a snap? That's so nice!

zacanger:
Tested on Ubuntu 18.04, snap version 2.39, and the install works, but the command installed is `entropic-ds.ds`. Is that intentional?

1stvamp:
Ah yes forgot to mention this, once the snap is official the `ds` alias can
be requested on the snapcraft forum, this is so snaps that override
executables in the Debian/Ubuntu archives, well, don't. 😁

zacanger:
Gotcha, that makes sense.

PR #272, opened 2019-06-14
Improved HACKING.md

niklabh:
# Change Type

* [ ] Feature
* [x] Chore
* [ ] Bug Fix

# Change Level

n/a

HACKING.md improvement for successful start.

- For successful docker-compose up .env-example should be copied to .env for all the services not just registry. npm i should be run in all the services.


zacanger:
None of those `/etc/hosts` changes should be necessary if you're using docker-compose. They talk to each other using those hostnames in the docker network, not the host network.

niklabh:
removed /etc/hosts

PR #275, opened 2019-06-14
Wait for DBs to start

CaptEmulation:
# Change Type

* [x] Feature
* [ ] Chore
* [ ] Bug Fix

# Change Level

* [ ] major
* [ ] minor
* [x] patch

When setting up locally via `docker-compose up` I had to try and try again otherwise containers that depended on redis and pg would fail to connect.  First I added some minor sleeps to the container commands and then later I came back and added a `retry_strategy` for redis and a query param to pg to extend the connection timeout.  The former is a code change and the latter was added to the relevant `.env-example`.

I am not adverse to adding tests or moving the redis stuff more into a config but these otherwise sensible values have been working well for me.

# Checklist

### DIFF
--- a/services/common/boltzmann/middleware/redis.js
+++ b/services/common/boltzmann/middleware/redis.js
@@ -26,7 +26,16 @@ function createRedisMW({
   redisURL = process.env.REDIS_URL || 'redis://localhost:6379'
 } = {}) {
   return next => {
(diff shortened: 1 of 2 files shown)


zacanger:
Nice! I ran into that when working on the docker-compose setup, but couldn't reproduce so I figured it was just an issue with my horribly slow machine.

PR #282, opened 2019-06-17
Project Cadence RFC

chrisdickinson:
Hi folks!

I'm opening this PR re: project cadence for a couple of reasons:

1. The project has been a lot to keep up with (thank you for all of your contributions!)
2. We've been moving fast to keep up with the contributions (and because we moved quickly, we made mistakes)
3. I've experienced the effect of a lack of healthy boundaries (esp. around availability & decision-making) while participating in other large OSS projects, and I want to call this out specifically before it becomes more of an issue.

This PR attempts to set boundaries by making the project cadence _explicit_ and calling out the need for more moderators.

As part of this effort, I'd like to welcome (& THANK) our new moderators with whom ceej and I will be sharing moderation duties:

- @varjmes
- @toddself 
- @aredridel

I've been continually impressed by how they've comported themselves on the tracker, discourse, and elsewhere, their ability to guide technical conversations to good outcomes, & the way they've kindly (but firmly!) course-corrected folks when the conversations get heated. (Again: thank y'all for agreeing to moderate! 💞)

* * *

The feedback I'm looking for in particular is:

- After reading this, do you feel like you understand when work is going to take place?
- Does this feel like a comfortable cadence?
- Is this fleshed out enough? What is missing?
- What is overspecified?

### DIFF
--- a/docs/RFCs/002-project-cadence.md
+++ b/docs/RFCs/002-project-cadence.md
@@ -1,0 +1,128 @@
+# Project Cadence
+
+A proposal for setting project work cadence
+
+- Version: DRAFT
+- Date: 06/17/2019
+- Authors
+  - Chris Dickinson <chris@neversaw.us>
+
+## problem statement
+
+Entropic is getting a lot of attention. This is **great**! However, because of
+the excitement, it means we have more issues and PRs to keep up with. We're
+attempting to interleave multiple different tasks: keeping up with ongoing
+conversations, moderation, code review, cross-maintainer coordination, and,
+well, our day jobs.
+
+We've also implicitly valued reaching "issue count 0" because we want
+collaborators to feel welcomed to the project. In order to reach that goal
+we've had to move quickly. As a result of moving quickly:
+
+- We're making mistakes,
+- Coordination suffers because we're _reacting_ to PRs instead of prioritizing
+  and discussing problems up front,
+- Quality of code review and moderation suffers because of context switching
+  across tasks.
+- We're not left with enough attention to spread across $dayjob, entropic, and
+  our personal lives.
+
+**So, let us take a step back and state the problem:**
+
+We wish to produce a healthy project that is capable of maintaining a federated
+package ecosystem over the long run.
+
+GitHub is a useful platform but it is incentivized to make people believe that
+project health is directly correlated to issue & PR count. **It's not.** GitHub is
+encouraging us to move fast at the expense of coordination.
+
+Context switching and lack of coordination are increasing friction, and the
+more the friction increases the faster we have to move to stay in the same
+place, making coordination worse. We are working in a reactive mode, and that
+is unlikely to produce a healthy project. We should do something about this!
+
+* * *
+
+## Antigoals
+
+We want to specifically avoid:
+
+- Slowing down too much and killing project momentum.
+- Overspecifying process. **Process is a system, and as a system it is subject to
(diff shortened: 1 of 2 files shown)


workingjubilee:
# Midweek, maybe?

Accidentally falling into Monday -> Friday patterns can result in a different kind of unhealthy where people overwork through the week and then fall over and die on the weekend instead of enjoying it. Or Monday -> Monday, where they hustle on the weekend to make stuff happen.

Not to say you can or should accommodate all schedules, nor can you necessarily automatically make everyone take breaks. But an approach that might encourage better-paced patterns might be scheduling anything important on a semi-weekly basis, like these proposed meetings, around Tue/Wed/Thu so that all possible weekdays have passed by essentially and it isn't starting/ending in sync with the traditional workweek.

## Do you just hate Mondays?

Maybe? Look, I may have a thing for lasagna but not that big of a thing.

## Is this sort of thing actually important?

This should definitely come way below "what times actually work at all". It's more of a tiebreaker thing, I think.

# Accessibility thoughts

I favor non-auditory communication a lot. It's easier to compose my thoughts. The project is in English, and that's a dealbreaker for a lot of people, but at least text can be read and parsed at leisure by people who aren't English first. And it already is possible to refer back to, as long as it is excerpted correctly. Whereas sound goes at varying speeds, is harder to excerpt and reference, and autotranscription via Google sucks.

Keeping a pace / to the minutes in text can be hard, though. Easier to self-distract. Tradeoffs exist.

non-binary:
Maybe not the place to post this so let me know if there's a better venue for it, but I'd be happy to assist with moderating among other things. I've been trying to figure out a way to get involved in the project but I'm a bit burned out from programming atm but excel at support and docs, which is what I do full-time at PSPDFKit :) Let me know if that's something I can help with.

docs/RFCs/002-project-cadence.md
```diff
@@ -0,0 +1,128 @@
...
+
+On this note: say "hello!" to [our new moderators](/README.md#moderators)! :tada:
+
+### 2nd step: "what's next?"
+
+We move back to a "what's next?" model for coordination and execution. This is
+modeled after how the project was developed before launch. 1-2 times a week --
+as we got tasks done or started to need to make decisions -- ceej and I would
+meet in person and go through an informal agenda of decisions that needed to be
+made and end with "what's next?" As in: what are the next most important tasks,
```
ceejbot: Needs a West Wing Jed Bartlet gif. Other than that, I like it.

strand: what's next?

toddself:
\Project meeting cadence is going to be important to keep up especially in the beginning when a large majority of :sparkles:decisions:sparkles: will need to be made.  I would say weekly until we've hashed out some of the underlying bits and then at least bi-weekly?

@workingjubilee had a great idea with tue/wed/thu; for some reason mid-week feels much less daunting than a friday (when i am frequently out my fucks to give for the week) and monday (when i'm trying to corral all those fucks so I can dole them out). Weekends are nearly impossible for me due to family commitments.

chrisdickinson:
Adding this to the merge queue since we've been doing this (roughly, minus televised meetings) in practice for a couple of weeks successfully. I've talked to some folks at the JS foundation about maybe providing tooling for casting Zoom to Youtube, which I'd like to switch to as soon as is convenient.

PR #290, opened 2019-06-20
doc: add LORE.md doc

chrisdickinson:
This adds a new document, `docs/LORE.md`, containing links to posts, writing, & tweets that inform the project's philosophy. (Thanks for the document naming suggestion, @iarna!) The goal is to keep a running document of important links instead of committing to writing up (& maintaining) a larger "project philosophy" post – this makes it easier to keep up to date, and thus more valuable for folks new to the project.

Had to resist attaching a gif of Brent Spiner. It was difficult, let me tell you.

What else should we add?

### DIFF
--- a/docs/LORE.md
+++ b/docs/LORE.md
@@ -1,0 +1,85 @@
+# LORE
+
+The intention of this document is to catalog a list of links to important or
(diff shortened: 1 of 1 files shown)


toddself:
image

PR #292, opened 2019-06-21
Added build step in circle CI

### LINKED ISSUE #281: Integration tests for docker-compose
We need tests to ensure changes are able to generate packages and that running docker-compose doesn't fail on these changes.

If anyone is interested in working on this that would be great!

nveenjain:
Added support to test docker-compose in circle CI
## Description

NA
## How to test
CircleCI will run build step automatically on this PR i guess..
## Checklist
NA


nveenjain:
Yeah, I too don't have much of docker expertise, just went through circleCI docs, would love any review and/or resources to follow.
Looking through the code of main repo now, so that I can contribute to this awesome idea more.

zacanger:
Related issue: #281

zacanger:
@nveenjain can you add `- build` to the `jobs` at the bottom of the file?

.circleci/config.yml
```diff
@@ -1,5 +1,17 @@
...
 jobs:
+  build:
+    docker:
+      - image: circleci/node:latest
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run: npm install
+      - run: 
+          name: Docker compose up
```
zacanger: This looks right to me! Would it be worth adding something to test that the containers stayed up? Maybe a `curl localhost:3000` to make sure the registry responds, or something like that?

zacanger:
I think that build got stuck in a loop. https://github.com/entropic-dev/entropic/pull/259 might be the fix, if you get up to date with upstream master I think that'll work.

nveenjain:
Ok thank, will try that!

nveenjain:
Looks like it is missing some environment variables, we could setup these environment variables in CI.
https://github.com/docker/compose/issues/4189#issuecomment-263458253
If they are already set, then https://github.com/docker/compose/issues/4189#issuecomment-456411061 might fix.. I'll try the second one now...

zacanger:
The issue here, I think, is actually that the docker-compose file points to `.env` files in the services, but those are usually created manually by users based on the `.env-example` files. Adding in a step before `docker-compose up` that copies each of those would probably work.

.circleci/config.yml
```diff
@@ -1,5 +1,21 @@
 version: 2
 jobs:
+  build:
```
zacanger: If you get rid of
```yaml
   docker:
      - image: circleci/node:latest
```

and replace it with
```yaml
  build:
    machine: true
```

that'll fix the containers starting up.

.circleci/config.yml
```diff
@@ -1,5 +1,21 @@
...
+      - setup_remote_docker
+      - run: npm install
+      - run: 
+          name: Docker compose up
+          command: |
+            set -ex
+            cp ./services/registry/.env-example ./services/registry/.env
+            cp ./services/web/.env-example ./services/web/.env
+            cp ./services/storage/.env-example ./services/storage/.env
+            docker-compose up -d
```
zacanger: I threw in a short sleep before the `curl` to give the containers time, and that worked on my build. I think 10 seconds would be plenty.

zacanger:
I debugged this over on my own Circle and had some suggestions.

zacanger:
This looks good! Should definitely help prevent breaking the dev setup in the future.

PR #299, opened 2019-06-23
fix(namespaces): change sort method in namespace model by order

ivandevp:

Fix

## Description

When requesting list of namespaces, an error is thrown:

Screen Shot 2019-06-23 at 6 27 17 PM

This is because the model is sorting the returning list by name but using an incorrect method from the ORM (`sort`). This PR changes the method to `order()` and the endpoint gets fixed.

## How to test

Request the `/v1/namespaces` without this change (it will return an error) and then with this change you will get the list of namespaces sorted by name.

## Checklist

### DIFF
--- a/services/storage/handlers/namespaces.js
+++ b/services/storage/handlers/namespaces.js
@@ -17,10 +17,7 @@ const exists = packageExists({
 });
 module.exports = [
   fork.get('/v1/namespaces', namespaces),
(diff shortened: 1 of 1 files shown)


chrisdickinson:
Merged in dffd49c7d1f50e1b4340567a2e321a396b0a2b3b, thank you!

PR #300, opened 2019-06-25
fix: package serialization changes

chrisdickinson:
Two changes contained:

1. Signatures should not be part of the package version content, because that's the data that's getting signed. Pluck them out!
2. Omit yanked versions from the top-level manifest. Once yanked, the package-version continues to be available at the content address and through the `/v1/packages/package/<pkg>/versions` endpoint. (Alternatively: we could add it to the versions map with `{yanked: <object id>}`.)

### DIFF
--- a/services/storage/models/package-version.js
+++ b/services/storage/models/package-version.js
@@ -51,6 +51,7 @@ module.exports = class PackageVersion {
     const {
       created,
       modified,
+      signatures,
       derivedFiles,
       ...content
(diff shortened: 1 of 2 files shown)


static-raccoon:
This pull request has been mentioned on **Entropic**. There might be relevant details there:

https://discourse.entropic.dev/t/testing-github-plugin-please-ignore/181/6

PR #303, opened 2019-06-27
nitpick(web): avoid ternary for plurals [WIP - do not merge yet] 

smikes:
Nitpick

## Description
It's much easier to internationalize when code is not mixed with formatting.  `count == 1 ? "" : "s"` in particular is hard to deal with: some languages have singular, dual, plural ; some languages have a special plural for numbers >= 5, etc.

## How to test
Not yet tested.

## Checklist

* [x] Added tests / did not decrease code coverage

### DIFF
--- a/services/web/handlers/auth.js
+++ b/services/web/handlers/auth.js
@@ -384,9 +384,9 @@ async function handleTokenAction(context) {
 
     context.session.set(
       'banner',
-      `Successfully deleted ${count} token${count === 1 ? '' : 's'}.`
+      `Successfully deleted tokens: ${count}.`
     );
-    context.logger.info(`${user.name} deleted a token`);
+    context.logger.info(`${user.name} deleted tokens: ${count}`);
     return response.redirect('/tokens');
   }
 }


ChrisBrownie55:
Note: [you can mark a PR as a Draft](https://github.blog/2019-02-14-introducing-draft-pull-requests/)

Instead of completely removing plural functionality, would it not be better to abstract this functionality to a separate function (avoids ternary but keeps plurals).

texh:
I feel like #115 is probably relevant here and some kind of general consensus on how i18n/l10n/etc will work needs to be met before going too far down the rabbit hole 🤷‍♂ .

chrisdickinson:
Thanks for the PR! We are thinking along the same lines, I believe. However, I'd prefer to replace these ternaries with `ngettext` (or similar) as part of an i18n pass through the codebase. If you (or anyone else) is interested in organizing efforts around that, that would be awesome. A good first step would be to open a discussion about what approach to use on [the Discourse](https://discourse.entropic.dev/). (I also understand that that's a much more involved effort, so do not worry if you don't have time for it!)

PR #315, opened 2019-07-14
fix(install): change hash for type on install script

juanarbol:

Fix

## Description

I changed `hash` for `type` because it was actually not falling to `elif` and `else` code blocks, just emit exception `hash: no such command: ...` and just put the `ds` on a variable because it was actually the same

Also I redirect `stderr` and `stdout` to `/dev/null` because `type` prints `command is /route` in `stdout`

## How to test
 Run `./misc/install.sh` on root folder
## Checklist

### DIFF
--- a/misc/install.sh
+++ b/misc/install.sh
@@ -3,10 +3,12 @@ set -e
 
 cd /tmp
 
-if hash curl 2>/dev/null; then
-  curl -SsL https://www.entropic.dev/ds-latest.tgz -o ds-latest.tgz
-elif hash wget 2>/dev/null; then
-  wget https://www.entropic.dev/ds-latest.tgz -o ds-latest.tgz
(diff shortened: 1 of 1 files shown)


zacanger:
Interesting that hash wasn't working for you, what system and shell are you on?

juanarbol:
Zsh on High Sierra

juanarbol:
Sorry, It actually works! It was a lapsus of me, now just put tarball resource on a variable, updated in [aaf12fa](https://github.com/entropic-dev/entropic/pull/315/commits/aaf12faae0de8c801223d57f2c8779301bc89550)

Sorry, again, my bad

zacanger:
Oh, cool! I was having a minor panic attack thinking I forgot something about how `hash` works :laughing:

juanarbol:
I do not know if this "contribution" has any real value, close this pull request in case of this is not really needed

PR #320, opened 2019-08-04
doc: fix name

frankgerhardt:


olingern:
I'm pretty sure `namespace` is the correct word to use here.

Relevant reading:
https://github.com/entropic-dev/entropic/tree/master/docs

frankgerhardt:
Ok, I agree. Invite confused me because it sounded like inviting a person or community but you seem to identify a person with a namespace.

What cleared it up is 
```
"name" - the canonical name of the package, as a string, including its namespace. e.g. "toddself@static-pkg.dev/packge2toml"
```
 on https://github.com/entropic-dev/entropic/blob/master/docs/rfcs/0002-manifest-toml.md

